PR #31, opened 2014-08-08
Adding ids to headers

srawlins:
I changed the (Setext)HeaderSyntax to generate an id for all headers generated, so that they can be deep-linked. This deep-linking can then also be used within a document using something like \[#this](text).

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -111,7 +111,6 @@ class BlockParser {
 }
 
 abstract class BlockSyntax {
-
   const BlockSyntax();
 
   /// Gets the regex used to identify the beginning of this block, if any.
@@ -144,6 +143,20 @@ abstract class BlockSyntax {
     if (parser.isDone) return true;
     return parser.blockSyntaxes.any((s) => s.canParse(parser) && s.canEndBlock);
   }
+
+  /// Generates a valid HTML anchor from the inner text of [element].
+  static String generateAnchorHash(Element element) =>
+      _concatenatedText(element)
+          .toLowerCase()
+          .trim()
+          .replaceFirst(new RegExp(r'^[^a-z]+'), '')
+          .replaceAll(new RegExp(r'[^a-z0-9 _-]'), '')
+          .replaceAll(new RegExp(r'\s'), '-');
+
+  /// Concatenates the text found in all the children of [element].
+  static String _concatenatedText(Element element) => element.children
+      .map((child) => (child is Text) ? child.text : _concatenatedText(child))
+      .join('');
 }
 
 class EmptyBlockSyntax extends BlockSyntax {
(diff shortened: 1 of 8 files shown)


srawlins:
Rebased.

munificent:
I definitely think this is useful, but I agree with @caffinatedmonkey that it's non-standard.

I think this should be done as an extension, but that would imply having some general architecture around extensions, which isn't quite there yet. I am interested in doing that, but I probably won't have time for a while.

**TL;DR: I like this, but it will be a while before I'm in a place to land it.**

munificent:
After you get the extension stuff figured out and landed, how about making this an opt-in extension?

srawlins:
@munificent this is ready for review now. Implemented as extensions.

lib/src/block_parser.dart
```diff
@@ -144,6 +143,20 @@ abstract class BlockSyntax {
     if (parser.isDone) return true;
     return parser.blockSyntaxes.any((s) => s.canParse(parser) && s.canEndBlock);
   }
+
+  /// Generate a valid HTML anchor from the inner text of [element].
+  static String generateAnchorHash(Element element) =>
+      _concatenatedText(element)
+          .toLowerCase()
+          .replaceAll(new RegExp(r'^[^a-z]+'), '')
```
munificent: Isn't this redundant with the next line?

lib/src/block_parser.dart
```diff
@@ -144,6 +143,20 @@ abstract class BlockSyntax {
...
+          .replaceAll(new RegExp(r'[^a-z0-9 _-]'), '')
+          .replaceAll(new RegExp(r'\s'), '-')
+          .replaceAll(new RegExp(r'-$'), '');
```
munificent: What about leading whitespace? Might be simplest just to trim the string before all the replaces.

lib/src/block_parser.dart
```diff
@@ -144,6 +143,20 @@ abstract class BlockSyntax {
...
+
+  /// Concatenate the text fonud in all the children of [element].
```
munificent: "found"

lib/src/html_renderer.dart
```diff
@@ -40,11 +42,13 @@ class HtmlRenderer implements NodeVisitor {
   static final _blockTags = new RegExp('blockquote|h1|h2|h3|h4|h5|h6|hr|p|pre');
 
   StringBuffer buffer;
+  LinkedHashSet<String> uniqueIds;
```
munificent: Why not just `Set`?

munificent:
Couple of suggestions and questions, but :+1:

lib/src/block_parser.dart
srawlins: This line kills all leading non-alphabetical characters. The next line kills all non alphanumeric+`[ _-]` characters, throughout. I don't think they're combinable.

lib/src/block_parser.dart
srawlins: Good call. Will do.

lib/src/block_parser.dart

lib/src/html_renderer.dart
srawlins: Done

srawlins:
Thanks @munificent !

PR #36, opened 2015-05-04
Apply hljs style to code block, not pre element

johnpryan:
The [highlighjs usage documentation]() encourages the `code` element to contain the css class for the language:

```
<pre><code class="html">...</code></pre>
```

This changes the converter to use that style instead of applying it to the `pre` element.

@dpeek

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -289,10 +289,11 @@ class FencedCodeBlockSyntax extends BlockSyntax {
     // Escape the code.
     final escaped = escapeHtml(childLines.join('\n'));
 
-    var element = new Element('pre', [new Element.text('code', escaped)]);
+    var codeElement = new Element.text('code', escaped);
     if (syntax != '') {
-      element.attributes['class'] = syntax;
+      codeElement.attributes['class'] = syntax;
     }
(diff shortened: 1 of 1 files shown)


munificent:
This would a be a breaking change for other consumers of the package that expect the current HTML output. :(

Ideally, this would be configurable in some way, but I don't know if I'll find the time to do that. Can you just massage the output HTML on your end to workaround this for now?

Also, file a bug if this is still important for you so I don't lose track of this. Thanks!

johnpryan:
I haven't decided how this should behave, I'll need to think about it some more.  I can work around this for now.  Thanks!

srawlins:
Arg. [CommonMark](http://spec.commonmark.org/0.22/#fenced-code-blocks) and [Pandoc](http://pandoc.org/README.html#fenced-code-blocks) disagree on what should have the class.

* CommonMark puts it on `<code>`, as in `<code class="language-dart">`.
* Pandoc puts it on `<pre>`, as in `<pre class="dart">`.
* GitHub puts it on a wrapper `<div>`, as in `<div class="highlight highlight-source-dart">`.

I'm leaning towards CommonMark, as configurability seems cumbersome. This would be a breaking change.

Issue #43, opened 2015-09-05
Formal specification for extensions beyond "standard" Markdown


srawlins:
This markdown package currently supports one non-standard syntax (the code fence), and there are open issues requesting support for more (see #18, #38, #41, and #31). It's of course up to the maintainers whether to include any more non-standard syntaxes beyond code fences, but the none of these issues has been outright rejected.

When syntax extensions are supported in a Markdown parser, they should be able to be toggled, to reduce surprises (see [Pandoc](http://pandoc.org/README.html#pandocs-markdown), [Redcarpet](https://github.com/vmg/redcarpet#and-its-like-really-simple-to-use), and others). This is the primary reason for having a formal specification: this package shouldn't format a user's Markdown strangely because it chose to include one certain non-standard syntax, and to always include it.

I propose a new argument to `markdownToHtml()` and `new Document()`, perhaps `List<String> extensions`. This argument defaults to `[]`. The different values might include:

* `"code_fences"` (already implemented)
* `"inline_html"` (#18)
* `"tables"` (#38)
* `"no_intra_word_emphasis"` (#41)
* `"header_ids"` (#31)

I realize that this package already supports user-written, _inline_ syntax extensions per-document, but supporting user-written _block_ syntaxes would be very hard, as they would want access to the different private patterns, etc (for example how ListSyntax uses all the other block syntax patterns to see whats up). So if this opens the door to non-standard block syntaxes like tables, and definition lists, then its a super win.

This new argument also eases some anxiety of adding a new syntax extension: it will always be safely disabled by default, and only enabled when the user chooses to use it.

If y'all like the idea, I'm happy to send a PR adding the argument, and disabling the FencedCodeBlockSyntax by default.

munificent:
Yeah, this is a good idea. There are other structural changes I'd like to do in the package, but my time to work on it is really limited right now. How about we leave this bug open for now and if I get some time to modernize the package overall, we'll do it then?

srawlins:
Hey @munificent ,

I just sent a PR for this formalization. If you have concrete ideas for the structural changes you want to make, could you open an issue? I might have time for it.

In any case, can we accept the formalization before the restructuring?

PR #44, opened 2015-09-07
Formalize extensions support; add inline HTML support

### LINKED ISSUE #18: Allow Inline HTML
Not sure if this is in scope, but markdown normally allows HTML tags. You can write something like this:

    This is **markdown'd**. But you can also <span color="red">do stuff manually</span>.

The span should be copied verbatim to the HTML output. In this implementation, though, all HTML tags are escaped (`<` becomes `&lt`).

srawlins:
Fixes #18 and #43 

I formalized support for Markdown extensions with an `extensions` optional parameter on `markdownToHtml()` and `new Document()`.

I then marked `fenced-code-blocks` as an extension (as they are), which serves as the block extension example, and added support for inline HTML (part of [CommonMark 0.22](http://spec.commonmark.org/0.22/#raw-html)), which serves as the inline extension example (and fixes #18).

The support for inline HTML is not fantastic. There seems to be a big desire for this package to be performant. Until benchmarking is more mature, I used a very very cheap (presumably), very very liberal HTML pattern. This extension is called `inline-html`.


lib/src/block_parser.dart
```diff
@@ -54,10 +54,34 @@ class BlockParser {
...
+    const OrderedListSyntax(),
+    const ParagraphSyntax()
+  ];
+
+  BlockParser(this.lines, this.document) {
+    _pos = 0;
+    if (document.extensions.contains('fenced-code-blocks')) {
+      blockSyntaxes.add(const FencedCodeBlockSyntax());
+    }
+    blockSyntaxes.addAll(standardBlockSyntaxes);
```
munificent: The way "extensions" or "plug-ins" work for InlineSyntaxes is that the parser pulls them from the document, which the public API can mutate. Outside code can just add in new objects that implement InlineSyntax.

I'm not sure if that's a great API or not—I lean towards not since it exposes an awful lot about how the parser is implemented—but I think it would be good for block syntax to be consistent with it.

You could either also add a `blockSyntaxes` field to Document, or remove `inlineSyntaxes` and use extensions for it.

lib/src/block_parser.dart
```diff
@@ -54,10 +54,34 @@ class BlockParser {
...
+    const BlockquoteSyntax(),
+    const HorizontalRuleSyntax(),
+    const UnorderedListSyntax(),
+    const OrderedListSyntax(),
+    const ParagraphSyntax()
+  ];
+
+  BlockParser(this.lines, this.document) {
+    _pos = 0;
+    if (document.extensions.contains('fenced-code-blocks')) {
```
munificent: Stringly typed APIs make me sad. How about an enum or make the extensions actual objects (i.e. similar to how the extension for a block syntax *is* the object that supports that syntax)?

test/markdown_test.dart
```diff
@@ -716,6 +722,40 @@ void main() {
...
+        '"markdown" within a tag is not parsed',
+        '''
+        Text <a href="_foo_">And "_foo_"</a>.
+        ''',
+        '''
+        <p>Text <a href="_foo_">And "<em>foo</em>"</a>.</p>
+        ''',
+        extensions: ['inline-html']);
+  });
+
```
munificent: I think this will all have to be rebased after I merged in @kevmoo's change that moves these tests out to data files. You'll probably want some way to specify which extensions a test data file requires.

lib/src/block_parser.dart
srawlins: I agree, I'm not crazy about exposing the innards. It's also unfortunate that the order of the syntaxes matters (maybe not so much with the block-level ones), because then the user can't insert their syntax in the middle.

I'll stick with how they work for InlineSyntaxes. Adding a `blockSyntaxes` field to Document. I can change the API back if we need to later.

lib/src/block_parser.dart
srawlins: Actual objects.

test/markdown_test.dart
srawlins: Done.

srawlins:
Thanks for the review, @munificent ! I've switched from strings to just objects, and I think its much cleaner now. Still not perfect, but at least consistent... let me know what you think.

kevmoo:
@srawlins would you want to rebase against master?

srawlins:
I think I did... recently

srawlins:
Or do you mean in the commit history? No I'll squash before merging. But maybe commit history is good for review for now.

lib/src/block_parser.dart
```diff
@@ -55,10 +55,32 @@ class BlockParser {
...
+    const CodeBlockSyntax(),
+    const BlockquoteSyntax(),
+    const HorizontalRuleSyntax(),
+    const UnorderedListSyntax(),
+    const OrderedListSyntax(),
+    const ParagraphSyntax()
+  ];
+
+  BlockParser(this.lines, this.document) {
+    _pos = 0;
```
munificent: Initialize this at the declaration?

test/extensions/inline_html.unit
```diff
@@ -0,0 +1,17 @@
+>>> wtihin a paragraph
```
munificent: "within" :)

munificent:
Two nits, but :+1:

munificent:
Oh, and don't forget to update the CHANGELOG, adjust the version in the pubspec if needed, etc.

lib/src/block_parser.dart
srawlins: Done

test/extensions/inline_html.unit
srawlins: done

srawlins:
Thanks for the review. This starts 0.9.0-dev, which will be unstable for a few more commits, methinks.

PR #48, opened 2015-09-12
Fix #47 by trimRight'ing

### LINKED ISSUE #47: Benchmark broken
When I run `dart benchmark/benchmark.dart`, it complains that the output isn't expected. When I diff it with `benchmark/output.html`, I get:

```diff
367c367,368
< </code></pre>
\ No newline at end of file
---
> </code></pre>
>
```

I think someone's editor killed the newline at the end or something (if I open `benchmark/output.html` in vim, and save and close, a newline is added, and git shows the file has changed).

To prevent annoying bugs like this, I suggest we `trimRight()` the actual and expected outputs.

srawlins:


munificent:
I probably accidentally deleted that newline. :)

I think the benchmark should be maximally precise, so I'll just fix the output file.

PR #49, opened 2015-09-13
Fix some bugs in the CodeSyntax implementation

srawlins:
This fixes some bugs in the CodeSyntax implementation. Namely:

1. Inline code delimited with single backticks should allow inner newlines.
2. Whitespace just inside backticks should be swallowed (not just inside double-backticks):

   ``before ` text ` after`` should render to `before <code>text</code> after`.
3. Numbers of backticks greater than 2 should be allowed.
4. Inline codes are not allowed to be adjacent to each other:

  ```before `foo `` bar` after``` should render to `before <code>foo `` bar</code> after`.

These are according to the rules of [CommonMark](http://spec.commonmark.org/0.22/#code-spans), and the canonical Markdown.pl implementation.

### DIFF
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -50,10 +50,7 @@ class InlineParser {
     // TODO(rnystrom): Underscores in the middle of a word should not be
     // parsed as emphasis like_in_this.
     new TagSyntax(r'_', tag: 'em'),
-    // Parse inline code within double backticks: "``code``".
-    new CodeSyntax(r'``\s?((?:.|\n)*?)\s?``'),
-    // Parse inline code within backticks: "`code`".
-    new CodeSyntax(r'`([^`]*)`')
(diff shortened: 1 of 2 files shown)


lib/src/inline_parser.dart
```diff
@@ -375,10 +375,41 @@ class ImageLinkSyntax extends LinkSyntax {
...
+  //   more.
+  //
+  // This conforms to the delimiters of inline code, both in Markdown.pl, and
+  // CommonMark.
+  static String _pattern = r'(`+(?!`))((?:.|\n)*?[^`])\1(?!`)';
+
+  CodeSyntax() : super(_pattern);
+
+  bool tryMatch(InlineParser parser) {
+    if (parser.pos > 0 && parser.source[parser.pos-1] == '`') {
```
munificent: Spaces around `-`. Or just run dartfmt on it. :)

PR #54, opened 2015-09-22
Fix [foo] (bar) bug, and [foo]() bug.

### LINKED ISSUE #53: Markdown converts [] space () to links
Take a look at 

https://github.com/dart-lang/sdk/blob/master/sdk/lib/developer/developer.dart#L40

Markdown converts it to links

https://api.dartlang.org/1.12.1/dart-developer/log.html

srawlins:
Fixes #53 

This also adds a little more conformance, making `[foo]()` render to a link.

### DIFF
--- a/test/original/inline_links.unit
+++ b/test/original/inline_links.unit
@@ -4,7 +4,7 @@ links [are](http://foo.com "woo") awesome
 <<<
 <p>links <a href="http://foo.com" title="woo">are</a> awesome</p>
 >>> no title
-links [are] (http://foo.com) awesome
+links [are](http://foo.com) awesome
 
 <<<
 <p>links <a href="http://foo.com">are</a> awesome</p>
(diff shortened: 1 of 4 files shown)


test/original/inline_links.unit
```diff
@@ -4,7 +4,7 @@ links [are](http://foo.com "woo") awesome
 <<<
 <p>links <a href="http://foo.com" title="woo">are</a> awesome</p>
 >>> no title
-links [are] (http://foo.com) awesome
+links [are](http://foo.com) awesome
```
munificent: Don't forget negative tests for `[inline] (space)` and `[reference] [space]` to validate that those do not become links.

munificent:
One thing then :+1:

test/original/inline_links.unit
srawlins: Done

srawlins:
Thanks for the review! :beer:

PR #55, opened 2015-09-22
Extension Sets

srawlins:
We should support CommonMark by default.

This does that, which means that calling `markdownToHtml()` as before will keep the fenced code blocks, which good for things like, all of the READMEs and dartdoc comments that are currently using fenced code blocks.

### DIFF
--- a/lib/src/document.dart
+++ b/lib/src/document.dart
@@ -2,22 +2,30 @@ library markdown.src.document;
 
 import 'ast.dart';
 import 'block_parser.dart';
+import 'extension_set.dart';
 import 'inline_parser.dart';
 
 /// Maintains the context needed to parse a Markdown document.
 class Document {
-  final Map<String, Link> refLinks;
+  final Map<String, Link> refLinks = {};
   List<BlockSyntax> blockSyntaxes;
   List<InlineSyntax> inlineSyntaxes;
+  ExtensionSet extensionSet;
   Resolver linkResolver;
   Resolver imageLinkResolver;
 
   Document(
-      {this.blockSyntaxes: const [],
-      this.inlineSyntaxes: const [],
+      {this.blockSyntaxes,
+      this.inlineSyntaxes,
+      extensionSet,
       this.linkResolver,
-      this.imageLinkResolver})
-      : refLinks = <String, Link>{};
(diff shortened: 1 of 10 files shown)


pubspec.yaml
```diff
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 description: A library for converting markdown to HTML.
 homepage: https://github.com/dart-lang/markdown
 environment:
-  sdk: '>=1.8.0 <2.0.0'
+  sdk: '>=1.12.0 <2.0.0'
```
srawlins: This is needed for `??=`. Is it crazy to bump the SDK for this reason?

munificent: Not crazy at all.

lib/src/document.dart
```diff
@@ -2,22 +2,30 @@ library markdown.src.document;
 
 import 'ast.dart';
 import 'block_parser.dart';
+import 'extension_set.dart';
 import 'inline_parser.dart';
 
 /// Maintains the context needed to parse a Markdown document.
 class Document {
-  final Map<String, Link> refLinks;
+  final Map<String, Link> refLinks = new Map<String, Link>();
```
munificent: `new Map<String, Link>()` -> `{}`

lib/src/document.dart
```diff
@@ -2,22 +2,30 @@ library markdown.src.document;
...
+      extensionSet,
       this.linkResolver,
-      this.imageLinkResolver})
-      : refLinks = <String, Link>{};
+      this.imageLinkResolver}) {
+    blockSyntaxes ??= [];
+    inlineSyntaxes ??= [];
+    extensionSet ??= ExtensionSet.commonMark;
+    blockSyntaxes.addAll(extensionSet.blockSyntaxes);
+    inlineSyntaxes.addAll(extensionSet.inlineSyntaxes);
```
munificent: You probably don't want to add redundant syntaxes, so it would be good to filter any existing ones out here.

munificent:
A couple of suggestions but :+1:

lib/src/document.dart
srawlins: Done.

srawlins:
Thanks @munificent

Issue #61, opened 2015-10-07
Need more help to add a block syntax extension


srawlins:
I'm trying to write a BlockSyntax for Description Lists ([pandoc description](http://pandoc.org/README.html#definition-lists), [PHP Markdown Extra description](https://michelf.ca/projects/php-markdown/extra/#def-list)), but I need some more helpers from BlockParser, and ListSyntax:

* BlockParser's `_pos` is (obviously) private, so the only way to peek at lines is with `get current` and `get next`. I need to see the line _2 lines_ ahead, so I propose a `String peek(int linesAhead)`.
* ListSyntax's `parse()` has a handy `blocksInList` object that should be a static final member of ListSyntax.

I intend to do the above.

PR #62, opened 2015-10-07
Provide a peek() for BlockParser, for extensions

srawlins:
This is for #61.

Just part 1.

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -91,6 +91,20 @@ class BlockParser {
     return lines[_pos + 1];
   }
 
+  /// Gets the line that is [linesAhead] lines ahead of the current one, or
+  /// `null` if there is none.
+  ///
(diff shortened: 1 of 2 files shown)


lib/src/block_parser.dart
```diff
@@ -91,6 +91,20 @@ class BlockParser {
...
   }
 
+  /// Gets the line that is [linesAhead] lines ahead of the current one, or
+  /// `null` if there is none.
+  ///
+  /// `peek(0)` is equivalent to [current].
+  ///
+  /// `peek(1)` is equivalent to [next].
+  String peek(int linesAhead) {
+    if (linesAhead == null || linesAhead < 0)
```
munificent: For what it's worth, we don't generally check for explicit null in most APIs.

munificent:
One comment but :+1:

PR #63, opened 2015-10-07
Rewrite ListSyntax.parse() to be simpler, and extract determineBlockItems()

### LINKED ISSUE #61: Need more help to add a block syntax extension

srawlins:
Fixes #61 

Essentially:

1. make `blocksInList` a public static List, so that extensions can use it.
2. Move all "is this list item a block item???" logic out of `parse()` into a new method, `determineBlockItems()`. Public so that extensions can use it.
3. Little cleanups like `pattern.firstMatch(...) != null` ==> `pattern.hasMatch(...)`

All tests continue to pass.

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -413,6 +413,16 @@ abstract class ListSyntax extends BlockSyntax {
 
   const ListSyntax();
 
+  /// A list of patterns that can start a valid block within a list item.
+  static final blocksInList = [
+    _blockquotePattern,
+    _headerPattern,
+    _hrPattern,
+    _indentPattern,
+    _ulPattern,
+    _olPattern
+  ];
+
   Node parse(BlockParser parser) {
     var items = <ListItem>[];
     var childLines = <String>[];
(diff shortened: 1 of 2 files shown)


srawlins:
@munificent , would you mind reviewing this refactor?

Heads up: GitHub's side-by-side diff is no good for this PR :tongue: better to just look at the new ListSyntax.parse() and ListSyntax.determineBlockItems().

lib/src/block_parser.dart
```diff
@@ -441,7 +451,29 @@ abstract class ListSyntax extends BlockSyntax {
...
+        itemNodes.add(new Element('li', contents));
+      }
+    }
 
+    return new Element(listTag, itemNodes);
+  }
+
+  /// Determines whether each item in [items] is a block item.
+  ///
+  /// Also removes any trailing empty lines and note which items are separated
```
munificent: "notes"

lib/src/block_parser.dart
```diff
@@ -484,56 +516,26 @@ abstract class ListSyntax extends BlockSyntax {
...
-        var children = parser.document.parseLines(item.lines);
-        itemNodes.add(new Element('li', children));
-      } else {
-        // Raw list item.
-        var contents = parser.document.parseInline(item.lines[0]);
-        itemNodes.add(new Element('li', contents));
-      }
-    }
+      // Items with more than one line are block items.
+      items[i].forceBlock = items[i].forceBlock || (items[i].lines.length > 1);
```
munificent: The parens aren't needed here.

munificent:
:+1:

lib/src/block_parser.dart
srawlins: done

lib/src/block_parser.dart
srawlins: done

srawlins:
Thanks @munificent !!

Issue #60, opened 2015-10-07
No way to hard newline


srawlins:
Similar to #30 , we should render escaped newlines as `<br />`:

```markdown
hello \
there
```

Issue #30, opened 2014-06-14
Two or more spaces at the end of the line should create line breaks


isoos:
See http://daringfireball.net/projects/markdown/syntax#p

> When you do want to insert a ``<br />`` break tag using Markdown, you end a line with two or more spaces, then type return.

PR #64, opened 2015-10-08
Fixing #30 and #60 by adding hard line break

srawlins:
The first regex change on inline_parser.dart:26 makes Markdown faster, while allowing the `r'  +\n'` to _not_ be the first TextSyntax.

Super oddly enough (to me), `\s*` was faster than`[ \t]*`, so I made the regex match multiple words in a row, separated by spaces or tabs, which makes it much faster again. So this PR doesn't slow down Markdown. Instead, it changes `benchmark.markdown` to go from 5.68 ms to 5.12 ms (9.9% faster).

### DIFF
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -23,7 +23,7 @@ class InlineParser {
 
     // TODO(amouravski): this regex will glom up any custom syntaxes unless
     // they're at the beginning.
-    new TextSyntax(r'\s*[A-Za-z0-9]+'),
+    new TextSyntax(r'[ \tA-Za-z0-9]*[A-Za-z0-9]'),
 
     // The real syntaxes.
     new AutolinkSyntax(),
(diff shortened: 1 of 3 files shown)


lib/src/inline_parser.dart
```diff
@@ -23,7 +23,7 @@ class InlineParser {
 
     // TODO(amouravski): this regex will glom up any custom syntaxes unless
     // they're at the beginning.
-    new TextSyntax(r'\s*[A-Za-z0-9]+'),
+    new TextSyntax(r'([ \t]*[A-Za-z0-9]+)+'),
```
ErikCorryGoogle: Would
[ \tA-Za-z0-9]*[A-Za-z0-9]
work too and be even faster?

munificent:
:+1:

lib/src/inline_parser.dart
srawlins: It is even faster! 4.86ms compared to 5.04ms.

srawlins:
Thanks @munificent and @ErikCorryGoogle !

PR #66, opened 2015-10-12
Fix newlines between link dest and title

### LINKED ISSUE #65: Whitespace should be allowed between a link's destination and title.
e.g.

```markdown
Links [are](http://foo.com
"woo") awesome.
```

srawlins:
Fixes #65.


munificent:
:+1:

PR #70, opened 2015-10-16
Adding a dartdoc-compare.rb script

srawlins:
Use this script to determine if changes to the markdown package change what dartdoc outputs.

Works on a regular package, and the SDK. Help text:

```
$ ruby tools/dartdoc-compare.rb --help
Usage: dartdoc-compare.rb [OPTIONS] <dart-package>
        --dartdoc-dir=DARTDOC_DIR    Directory of the dartdoc package
        --before=BEFORE              Markdown package 'before' ref
        --after=AFTER                Markdown package 'after' ref (default: HEAD)
        --sdk                        Is the package the SDK?
```

### DIFF
--- a/tools/dartdoc-compare.rb
+++ b/tools/dartdoc-compare.rb
@@ -1,0 +1,89 @@
+require 'optparse'
+require 'tmpdir'
+require 'yaml'
+
+options = {markdown_after: 'HEAD', sdk: false}
+
+OptionParser.new do |opts|
+  opts.banner = "Usage: dartdoc-compare.rb [OPTIONS] <dart-package>"
+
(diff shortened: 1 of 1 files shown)


sethladd:
This is really cool, but could it be a Dart script? :)

srawlins:
I'd like it to be a Dart script to. Unfortunately, no Yaml-writing that I could find. dart-lang/yaml#2. So it's Ruby for now :)

I think that was the only blocker though.

munificent:
> Unfortunately, no Yaml-writing that I could find.

You can just write it as JSON. YAML is a superset of JSON.

srawlins:
Superseded by #81

PR #71, opened 2015-10-23
Fix over-eager word syntax

srawlins:
When writing a custom syntax for, say, a common way that bugs are written, like "bug#123", I wanted to write a custom parser for  auto linking that.

Unfortunately, the first, very hungry syntax eats up 'b', 'u', and 'g', before bailing. Then the custom syntax is left with "#123", and catches nothing.

This makes the inline parser less eager, if custom syntaxes appear to exist. I would make it _always_ less eager, except this less eager parser take **3x** as long in `benchmark.dart` :(

Tests pass.

### DIFF
--- a/lib/src/document.dart
+++ b/lib/src/document.dart
@@ -20,13 +20,13 @@ class Document {
       ExtensionSet extensionSet,
       this.linkResolver,
       this.imageLinkResolver}) {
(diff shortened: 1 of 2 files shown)


kevmoo:
@srawlins I may have messed up your pull request. Rebase?

srawlins:
K I think it's ready again.

Issue #74, opened 2015-10-29
Nested bullets aren't rendering correctly


sethladd:
From this file: https://github.com/dart-lang/shelf/blob/master/README.md

At the bottom, on Github, we see this:

![screen shot 2015-10-29 at 8 03 48 am](https://cloud.githubusercontent.com/assets/5479/10822271/b03e66f0-7e13-11e5-88ea-b7255b1c1aa7.png)

When dartdoc, which uses this markdown package, renders that file, it looks like this:

![screen shot 2015-10-29 at 8 04 57 am](https://cloud.githubusercontent.com/assets/5479/10822296/c42f8b6c-7e13-11e5-902d-9f12b3cce6f3.png)

(note: I didnt verify that the bug is _definitely_ with markdown package. dartdoc could be doing something weird here)

sethladd:
cc @keertip

srawlins:
:frowning: legit.

srawlins:
OK @sethladd, I investigated this a bit; we're using the 4-space rule, and a strictness in lists, so:

```markdown
* one
  *  nested
```

will NOT nest. You need 4 spaces before the nested `*`. Also:

```markdown
* one
    * nested
```

will NOT nest. You need a blank line before lists:

```markdown
* one

    * two
```

should nest. The example in @munificent 's [Effective Dart: Documentation](https://www.dartlang.org/effective-dart/documentation/#markdown) correctly shows how to nest lists.

srawlins:
One might also consider this a dup of #14

sethladd:
Should we match GitHub here? Github's MD seems to be a defacto standard.

kwalrath:
In my markdown experience, I expect (through painful experience) to need a blank line above the first bullet, but none in the rest of the list.

Requiring blank lines within lists feels broken.

sethladd:
> Requiring blank lines within lists feels broken.

Agreed.

munificent:
> Should we match GitHub here? Github's MD seems to be a defacto standard.

I think we should generally aim to match CommonMark but if there are places where that diverges from GitHub, we'll want to think long and hard about which side to lean towards.

> Requiring blank lines within lists feels broken.

Here, it's only needed to start the nested list, but, yes, I agree. Markdown is actually really weird here. The presence of newlines affects the resulting output. If you have:

```markdown
* A

* B

* C
```

Then you get an unordered list where each item is a paragraph. If you do:

```markdown
* A
* B
* C
```

[...]

PR #75, opened 2015-11-05
Add a nested list test

srawlins:
This "documents" #74. I just saw this is also documented at [Effective Dart: Documentation](https://www.dartlang.org/effective-dart/documentation/#markdown).


munificent:
:+1:

PR #80, opened 2015-11-17
Follow CommonMark convention for language in fenced code blocks

kevmoo:
Improves CommonMark fenced code test results from 13/27 to 16/27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.10.0
+
+* BREAKING: Now following the CommonMark spec for fenced code blocks.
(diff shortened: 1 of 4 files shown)


srawlins:
Love it. Does this new format work with the syntax highlighting we do at pub.dartlang? dartdocs.org? api.dartlang.org?

kevmoo:
@srawlins `prettify`, which is used by dartdoc "does the right thing" with these updates. Not sure about the pub site – hence the breaking change.

srawlins:
Okie doke, LGTM.

Issue #51, opened 2015-09-15
CommonMark compliance


Scorpiion:
Hi,

Is there a plan for this package to become "CommonMark compliant" or is the intention to follow some specific (or it's own) flavor of Markdown?

I have been using a different Dart package, md_proc, simply because it claims to be CommonMark compliant. Not sure how these two package differs, but if this package also intends to become CommonMark compliant maybe there are some "synergies" to be found between the two: 
https://github.com/dikmax/md_proc

srawlins:
My intention would be for this package to be come CommonMark compliant, and run it against the CommonMark specs.

md_proc looks like a very full-featured Markdown parser! Very cool project. Not sure what opportunities for "synergies" there would be though...

Scorpiion:
Okay, yes I would like that too. I think it's nice if the parser would follow a clear spec.

Yeah I know, but it's a good reference at least. Maybe some of the tests against the spec could be reused or base new tests upon them to save at least a little bit of time.

kevmoo:
FYI – I've added some tooling to at least validate this impl against CommonMark - https://github.com/dart-lang/markdown/commit/a7db5a63b6b6fd640fb7fd6ce49138df16851f16

We fail a lot of these – but at least we have a way to track our compliance

mehaase:
Comparing test cases as strings won't work because the test cases contain HTML attributes in arbitrary orders, e.g.: 

    {
      "section": "Images",
      "end_line": 7244,
      "markdown": "![foo](/url \"title\")\n",
      "start_line": 7240,
      "html": "<p><img src=\"/url\" alt=\"foo\" title=\"title\" /></p>\n",
      "example": 521
    },

The `HtmlRenderer` in this library sorts attribute names alphabetically, so it will never produce "src" before "alt".

kevmoo:
Would it be better if `HtmlRender` *didn't* sort the elements?

mehaase:
There's probably no good reason to sort attributes, but the unit tests currently depend on attribute order so additional work is needed if attributes are not sorted.

kevmoo:
@mehaase I just went through to see if sorting affects any of the common-mark tests. I doesn't seem that it does.

If it comes up as a spec compliance thing, I'd consider it – otherwise I'd like to avoid adding noise to the implementation.

mehaase:
> @mehaase do you have an example where a common-mark test is failing just because of the ordering of attributes?

[...]

PR #84, opened 2016-02-17
Improve dart-lang/markdown#51.

mehaase:
Improves dart-lang/markdown#51.

First, put a try/catch around attempting to parse Markdown.
Failure to parse should count as a failed test case, but
should not abort testing altogether.

Second, implement a primitive HTML comparison function that
is not sensitive to attribute order, meaningless whitespace,
etc.

The second change alone boosts CommonMark score from 232/599
to 383/599.


tool/common_mark_stats.dart
```diff
@@ -62,6 +72,60 @@ void main() {
       '– ${pct}%');
 }
 
+/// Compare two DOM trees for equality.
+bool compareHtml(List<Element> expectedEls, List<Element> actualEls) {
```
kevmoo: don't use abbreviations here –> `expectedElements`, etc

tool/common_mark_stats.dart
```diff
@@ -62,6 +72,60 @@ void main() {
...
+  if (expectedEls.length != actualEls.length) {
+    return false;
+  }
+
+  for (int childNum in new List.generate(expectedEls.length, (i) => i)) {
```
kevmoo: don't needlessly type `var childNum` etc...

Also ...just use `for (var childNum = 0; childNum < expectedElement.length; childNum++)`

tool/common_mark_stats.dart
```diff
@@ -62,6 +72,60 @@ void main() {
...
+    if (expected is Element) {
+      if (expected.localName != actual.localName) {
+        return false;
+      }
+
+      if (expected.attributes.length != actual.attributes.length) {
+        return false;
+      }
+
+      List expectedAttrKeys = new List.from(expected.attributes.keys);
```
kevmoo: `var`

Also: `expected.attributes.keys.toList()`

tool/common_mark_stats.dart
```diff
@@ -62,6 +72,60 @@ void main() {
...
+      expectedAttrKeys.sort();
+
+      List actualAttrKeys = new List.from(actual.attributes.keys);
+      actualAttrKeys.sort();
+
+      for (int attrNum in new List.generate(actualAttrKeys.length, (i) => i)) {
```
kevmoo: same as above

tool/common_mark_stats.dart
```diff
@@ -62,6 +72,60 @@ void main() {
...
+        }
+
+        if (expected.attributes[expectedAttrKey] !=
+            actual.attributes.keys[actualAttrKey]) {
+          return false;
+        }
+      }
+    }
+
+    bool childrenEqual = compareHtml(expected.children, actual.children);
```
kevmoo: var

kevmoo:
@mehaase do you have an example where a common-mark test is failing just because of the ordering of attributes?

mehaase:
Just pushed an update that address your feedback.

mehaase:
> @mehaase do you have an example where a common-mark test is failing just because of the ordering of attributes?

*I answered this question in a separate thread, but I realize that it's hard to keep track of the different threads, so I'm reposting that answer here.*

Test case 521 is failing because of the order of attributes. Here are the expected and actual outputs for that test case as it stands on `master` right now.

Expected:

    <p>foo</p>

Actual:

    <p>foo</p>

There are other test cases that rely on attribute order. Most of them are anchors that happen to have attributes in alphabetical order (`href` before `title`). The image test cases have `src` before `alt`, which will definitely cause test failures, but nobody noticed before because the image test cases were totally broken for other reasons.

The far, far more common issue, however, is insignificant whitespace. For example, test case 4:

Expected:

    <ul>
    <li>
    <p>foo</p>
    <p>bar</p>
    </li>
    </ul>

Actual:

    <ul><li>
    <p>foo</p>
    <p>bar</p></li></ul>

It's the exact same DOM. The whitespace between `<ul>` and `<li>`, for example, is insignificant. 

**Over 150 CommonMark tests fail due to insignificant differences in the markup.** Moreover, you'll *never get these tests right* unless you normalize the HTML output somehow. 

For example, jgm's repo (which is where you got the test cases from, right?) performs [a sophisticated normalization process](https://github.com/jgm/CommonMark/blob/master/test/normalize.py) so that HTML can be compared for significant differences.

srawlins:
Hi @mehaase, sorry for the crazy long wait, can you fix merge conflicts?

srawlins:
This is a great change. I can confirm that this patch improves our compliance from **237 of 613 -- 38.7% total** to **400 of 613 -- 65.3% total**.

@mehaase, if you can fix the conflicts surrounding the change from `validCount` to `nestedMap` (should be pretty simple), I can merge this.

mehaase:
Thanks @srawlins. I fixed the merge conflicts.

srawlins:
All tests pass, CommonMark coverage dramatically improved. Thanks!

mehaase:
Thank you!

Issue #83, opened 2016-02-15
Images not rendered correctly


mehaase:
#19 was closed, but why? Image rendering is still totally broken. An inline image without a hyperlink is rendered with a hyperlink, and an inline image with a hyperlink is completely corrupted, producing ill-formed HTML markup, illegal URLs, and orphaned characters.

## Example 1 - image without hyperlink:

    ![a robot](http://www.littlerobotsoundfactory.com/img/8-bit_1.png)

This renders as:

    <p>
      <a href="http://www.littlerobotsoundfactory.com/img/8-bit_1.png">
        <img alt="a robot" src="http://www.littlerobotsoundfactory.com/img/8-bit_1.png">
        </img>
      </a>
    </p>

(I added newlines and indentation to clarify structure; the actual rendering has no whitespace between tags.) In addition to the superfluous `<a>` element, note the illegal `</img>` tag.

The correct rendering is:

    <p>
        <img alt="a robot" src="http://www.littlerobotsoundfactory.com/img/8-bit_1.png">
    </p>

## Example 2 - image with hyperlink:

This image *should* be hyperlinked to robotstxt.org:

    [![a robot](http://www.littlerobotsoundfactory.com/img/8-bit_1.png)](http://www.robotstxt.org/)

But it unfortunately renders like this:

    <p>
      [
      <a href="http://www.littlerobotsoundfactory.com/img/8-bit_1.png)](http://www.robotstxt.org/">
        <img alt="a robot" src="http://www.littlerobotsoundfactory.com/img/8-bit_1.png)](http://www.robotstxt.org/">
        </img>
      </a>
    </p>

(Again, whitespace added to clarify structure.) This rendering is truly bizarre and broken. First, there is an orphaned `[` as the first child of `<p>`. Then, the `href` of the anchor is totally wrong: `http://www.littlerobotsoundfactory.com/img/8-bit_1.png)](http://www.robotstxt.org/`. It combines the two URLs in an unexpected way and includes characters that are not legal in a URL. The `src` of the image is set to the same string as `href`, thus also totally wrong. Finally, there's the illegal `</img>` closing tag again.

## Standard

The CommonMark reference implementation does not hyperlink an inline image, nor does Gruber's Markdown. Why then does Dart Markdown choose to do this? Markdown syntax already supports putting an image inside of a link, so it doesn't make sense to have 2 syntaxes for hyperlinking an image and 0 syntaxes for an image without hyperlink.

[...]

PR #85, opened 2016-02-17
Fixes dart-lang/markdown#83.

mehaase:
Fixes dart-lang/markdown#83.

An inline image is now rendered correctly, sans anchor element,
and not emitting a close tag for the image element. Unit tests
have been updated to match correct rendering.

An inline image may also be placed inside the content of a link,
and this case is also rendered correctly. I added three new unit
tests for this.

This *would* boost the CommonMark test for images from 1/22 to
9/22 and the test for links from 33/79 to 53/79, except the
CommonMark tests are sensitive to attribute order. (See separate
pull request on dart-lang/markdown#51.)


srawlins:
This looks fantastic. I did not realize what our current image code was doing. Thanks for the hard work, @mehaase!

Issue #86, opened 2016-02-17
Line breaks not parsed correctly


mehaase:
Line breaks (e.g. the two spaces at the end of a line) are not parsed correctly. Line breaks should emit a `<br>` tag, but `InlineParser` emits it as a Text node, not an element node. This results in a correct rendering when using the supplied `HtmlRenderer`, but makes other `NodeVistor` implementations useless. 

For example, consider this node visitor, which should print an element tree:

    class AstPrinter implements md.NodeVisitor {
        int indent = '';

        void visitElementAfter(md.Element element) {
            print('${indent}leaving: ${element.tag}');
            indent = indent.substring(0, indent.length-2);
        }

        bool visitElementBefore(md.Element element) {
            indent += '  ';

            if (element.isEmpty) {
                print('${indent}void: ${element.tag}');
                indent = indent.substring(0, indent.length-2);
                return false;
            } else {
                print('${indent}entering: ${element.tag}');
                return true;
            }
        }

        void visitText(md.Text node) {
            indent += '  ';
            List<String> lines = node.text.replaceAll('\r\n', '\n').split('\n');
            for (String line in lines) {
                print('${indent}TEXT> ${line}');
            }
            indent = indent.substring(0, indent.length-2);
        }
    }

The driver code looks like this:

    List<String> lines = ['*hello* **world**!', '', '> this is a quote', 'good bye!'];
    md.Document doc = new md.Document();
    AstPrinter printer = new AstPrinter();

    for (md.Node node in doc.parseLines(lines)) {
        node.accept(printer);
    }

(I've imported markdown as `md` so it doesn't conflict with `core:html`.) 

The output of this example is:

    entering: p
      entering: em
        TEXT> hello
      leaving: em
      TEXT>  
      entering: strong
        TEXT> world
      leaving: strong
      TEXT> !
    leaving: p
    entering: blockquote
      entering: p
        TEXT> this is a quote
      leaving: p
    leaving: blockquote
    entering: p
      TEXT> good bye!
    leaving: p

(Side note, since there are no unit tests for proper AST construction, this `AstPrinter` would be a handy class to include in the Markdown library for debugging purposes.)

What tree is printed when a linebreak is parsed? Here's a test:

    List<String> lines = ['this is a test  ', 'this is only a test'];

The result: 

    entering: p
      TEXT> this is a test
      TEXT> <br />
      TEXT> this is only a test
    leaving: p

[...]

PR #87, opened 2016-02-17
Fixes dart-lang/markdown#86.

mehaase:
Fixes dart-lang/markdown#86.

Linebreaks should not be treated as `TextSyntax` but should be
a subclass of `InlineSyntax`. This also requires a small
modification to HtmlRenderer to put a \n after each &lt;br /&gt;;
this makes all the unit tests pass.

### DIFF
--- a/lib/src/html_renderer.dart
+++ b/lib/src/html_renderer.dart
@@ -85,6 +85,11 @@ class HtmlRenderer implements NodeVisitor {
     if (element.isEmpty) {
       // Empty element like <hr/>.
       buffer.write(' />');
+
+      if (element.tag == 'br') {
+        buffer.write('\n');
+      }
+
       return false;
     } else {
       buffer.write('>');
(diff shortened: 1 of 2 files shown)


kevmoo:
Can we have a test that validates this output? I'm a bit of a loss how what affect the change has...

mehaase:
It's not really a test of the output, it's a test of the AST.

Before:

    entering: p
      TEXT> this is a test
      TEXT> <br />
      TEXT> this is only a test
    leaving: p

After:

    entering: p
      TEXT> this is a test
      void: br
      TEXT> this is only a test
    leaving: p

`br` is now a real element — a node in the AST — not just a line of text inside a `Text` node.

All of the existing unit tests are based on rendered output, so intermediate representations (AST) are not tested at all.

In my application, I'm not generating HTML. I am walking over the AST and creating corresponding DOM nodes to insert directly into my document. I was flummoxed when I started creating nodes that had the text `<br />` in them wherever an actual line break element was supposed to be.

mehaase:
Re-opening because this change was [reverted on master](https://github.com/dart-lang/markdown/issues/86#issuecomment-185844585).

mehaase:
@kevmoo @srawlins Can you please look at this one again? It was merged and then reverted. Kevin thought that it broke the CommonMark tests, but actually it works fine with the other PR that you just merged.

PR #88, opened 2016-02-17
Latest spec

kevmoo:
I've pulled in the latest common_mark spec and I'm saving the output of the tool (both a full diff and summary) in the tool directory so we can track how one-off changes affect compliance.

Nice?


srawlins:
LGTM

PR #94, opened 2016-05-27
Support shortuct reference links and images

srawlins:
This adds support for:

```markdown
[foo] and ![bar].

[foo]: http://example.com
[bar]: http://example.com/bar.png
```

CommonMark: http://spec.commonmark.org/0.25/#shortcut-reference-link

This bumps our CommonMark compliance from 401 to 428. Woohoo!


munificent:
Awesome! 👍

PR #95, opened 2016-05-27
A code blocks cannot interrupt a paragraph

srawlins:
From CommonMark: http://spec.commonmark.org/0.25/#indented-code-blocks

> An indented code block cannot interrupt a paragraph.
> (This allows hanging indents and the like.)

This bumps our CommonMark compliance from 401 to 407.


munificent:
👍

PR #97, opened 2016-05-31
Improve the block HTML identification

srawlins:
Better CommonMark [HTML Blocks](http://spec.commonmark.org/0.24/#html-blocks). This makes us compliant with 29 more CommonMark 0.24 tests.

All of the existing tests pass.

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -34,10 +34,6 @@ final _codePattern = new RegExp(r'^[ ]{0,3}(`{3,}|~{3,})(.*)$');
 /// SETEXT should win.
 final _hrPattern = new RegExp(r'^ {0,3}([-*_]) *\1 *\1(?:\1| )*$');
 
-/// Really hacky way to detect block-level embedded HTML. Just looks for
-/// "<somename".
-final _htmlPattern = new RegExp(r'^<[ ]*\w+[ >]');
-
 /// A line starting with one of these markers: `-`, `*`, `+`. May have up to
 /// three leading spaces before the marker and any number of spaces or tabs
 /// after.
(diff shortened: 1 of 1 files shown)


lib/src/block_parser.dart
```diff
@@ -65,9 +61,17 @@ class BlockParser {
...
 
   /// The collection of built-in block parsers.
-  final List<BlockSyntax> standardBlockSyntaxes = const [
+  final List<BlockSyntax> standardBlockSyntaxes = [
     const EmptyBlockSyntax(),
-    const BlockHtmlSyntax(),
+    const BlockTagBlockHtmlSyntax(),
+    new LongBlockHtmlSyntax(r'^ {0,3}<pre(?:\s|>|$)', '</pre>'),
+    new LongBlockHtmlSyntax(r'^ {0,3}<script(?:\s|>|$)', '</script>'),
+    new LongBlockHtmlSyntax(r'^ {0,3}<style(?:\s|>|$)', '</style>'),
```
munificent: How come these don't use BlockTagBlockHtmlSyntax? Can you add a comment explaining?

lib/src/block_parser.dart
```diff
@@ -413,6 +424,40 @@ class BlockHtmlSyntax extends BlockSyntax {
   }
 }
 
+class OtherTagBlockHtmlSyntax extends BlockTagBlockHtmlSyntax {
+  bool get canEndBlock => false;
+
+  // Really hacky way to detect "other" HTML.
+  RegExp get pattern => new RegExp(r'^ {0,3}</?\w+(?:>|\s+[^>]*>)\s*$');
```
munificent: This RegExp hurts my mind. How about a comment explaining it?

lib/src/block_parser.dart
srawlins: Sorry, added a comment to LongBlockHtmlSyntax:

> A BlockHtmlSyntax that has a specific [endPattern].
>
> In practice this means that the syntax dominates; it is allowed to eat
> many lines, including blank lines, before matching its [endPattern].

munificent:
👍

PR #102, opened 2016-07-05
Fix various list bugs

srawlins:
This fixes various list bugs, bringing our CommonMark compliance up from 465 (75.5%) to 483 (78.4%).

* Nested lists were only enabled with 3 spaces or a tab. Now you need to indent as far as the content of the parent item.
* Blank list items are allowed.
* Two blank lines ends a list.
* List items that start with an indented code block work ok.

PR #103, opened 2016-07-05
Using CommonMark's 'tight' and 'loose' lists

srawlins:
This switches our definition of "block-level" list items and "inline" list items to CommonMark's [loose and tight lists](http://spec.commonmark.org/0.25/#lists).

It improves our CommonMark score from 465 to 468, but I think that this improves the score found in #102.


srawlins:
CC @kevmoo @munificent

munificent:
:+1:

PR #105, opened 2016-07-07
Find and fix a gnarly oversight with using the wrong 'Element' class

kevmoo:
Thanks, strong-mode


srawlins:
Whoa, wait. Can you explain this? Our spec script was broken so that we thought we were passing more specs than we actually were?

kevmoo:
> Our spec script was broken so that we thought we were passing more specs than we actually were?

Yup.

srawlins:
Frowny face. 😦 . Okie doke though. Thanks for finding and fixing!

PR #106, opened 2016-07-07
Better tools

kevmoo:
Now it should be really easy to update the common mark stats on every CL


srawlins:
Thanks much, this is handy!

PR #108, opened 2016-07-08
Print caught errors in CommonMark script

srawlins:


tool/common_mark_stats.dart
```diff
@@ -75,15 +75,26 @@ main(List<String> args) async {
...
       var nestedMap =
           scores.putIfAbsent(section, () => new SplayTreeMap<int, bool>());
+      var expected = parseFragment(e.html);
 
       try {
         output = markdownToHtml(e.markdown);
-      } catch (exc) {
+      } catch (exc, stackTrace) {
         nestedMap[e.example] = false;
+        if (verbose) {
```
kevmoo: Seems like a *lot* of duplication with the code starting at line 101 – could this be factored into some shared code? Even creating a constant for the spec link would be a start

srawlins: Done

PR #110, opened 2016-07-12
Overhaul link reference definitions.

srawlins:
The new process is similar to CommonMark.js, where reference link definitions
are parsed with block nodes, and inline nodes are parsed after all block nodes
have been parsed. The tests highlight the big new features:

* Link reference definitions can now span multiple lines. This is a big one, because this is common:

  ```markdown
  [the text from a paragraph where this link is]:
  http://a.really.long.url/com/that-should-go-on-its-own-line
  ```
* Link reference definitions can now be found nested somewhere:

  ```markdown
  * like in list [items].

    [items]: http://foo.com

    > and [blockquotes].
    >
    > [blockquotes]: http://bar.com
  ```

And as can be seen by `tool/common_mark_stats.txt`, this adds 10 more compliant CommonMark specs.


tool/common_mark_stats.json
```diff
@@ -461,15 +461,15 @@
...
+  "510": true,
   "511": true,
   "512": false,
   "513": true,
   "514": true,
   "515": true,
   "516": true,
   "517": true,
-  "518": true,
+  "518": false,
```
kevmoo: @srawlins A regression?

lib/src/block_parser.dart
```diff
@@ -728,7 +730,137 @@ class ParagraphSyntax extends BlockSyntax {
...
+      // Paragraph consisted solely of reference link definitions.
+      return new Text('');
+    } else {
+      var contents = new UnparsedContent(paragraphLines.join('\n'));
+      return new Element('p', [contents]);
+    }
+  }
+
+  // Extract reference link definitions from the front of the paragraph, and
+  // return the remaining paragraph lines.
```
munificent: Nit: `///` for doc comment.

lib/src/block_parser.dart
```diff
@@ -728,7 +730,137 @@ class ParagraphSyntax extends BlockSyntax {
...
+        r'''\[([^\]]+)\]:\s+(?:<(\S+)>|(\S+))\s*'''
+        // Title in double or single quotes, or parens.
+        r'''("[^"]+"|'[^']+'|\([^)]+\)|)\s*$''',
+        multiLine: true);
+    var match = pattern.firstMatch(contents);
+    if (match == null) {
+      // Not a reference link definition.
+      return false;
+    }
+    if (match[0].length < contents.length) {
```
munificent: Doesn't the "$" in the pattern cover this?

lib/src/ast.dart
```diff
@@ -65,6 +65,13 @@ class Text implements Node {
   String get textContent => text;
 }
```
munificent: Doc comment?

lib/src/block_parser.dart
```diff
@@ -728,7 +730,137 @@ class ParagraphSyntax extends BlockSyntax {
...
+    } else {
+      var contents = new UnparsedContent(paragraphLines.join('\n'));
+      return new Element('p', [contents]);
+    }
+  }
+
+  // Extract reference link definitions from the front of the paragraph, and
+  // return the remaining paragraph lines.
+  List<String> _extractReflinkDefinitions(
+      BlockParser parser, List<String> lines) {
```
munificent: This function is pretty hairy. (Fun fact: this is the first time I've seen continue to a label in real Dart code). My intuition is that there should be a simpler way to express this.

Would this idea work?

Run through the list of lines looking for reflink definitions. When you find one, extract it and replace those elements in the list with `null`. Then, at the end, filter out the null elements. That way you don't keep creating and throwing away new lists.

Would that work?

tool/common_mark_stats.json
srawlins: Fixed.

lib/src/block_parser.dart

lib/src/block_parser.dart
srawlins: Unfortunately no. In this function, \[contents] can be multiple lines joined together (newline characters intact), and `$` matches the end of the line, rather than the end of the string, so this check is needed for examples [170](http://spec.commonmark.org/0.25/#example-170) and [172](http://spec.commonmark.org/0.25/#example-172).

lib/src/ast.dart
srawlins: Done.

lib/src/block_parser.dart
srawlins: This function _is_ hairy 😟  It took me a few re-writes to even get this 😛 .

However, I think your suggestion won't work, for one annoying reason: Link reference definitions, which look like:

> [label]: http://url "title of many words"

are allowed to not have a title, and they're allowed to have a title on the line following the URL, and the title is allowed to be multiline, so the following are all link reference definitions:

> [label]: http://url
>
> [label]: http://url
> "title"
>
> [label]: http://url
> "title
> with many
> words"

So (I think), we have to do this nutty thing that I'm doing, first sectioning the "paragraph" into multiple possible link reference definitions, then looking closely at each one to see a) if it even is a real link reference definition, and b) what's the "longest" link reference definition to be found.

So I don't know how to parse certain examples like [157](http://spec.commonmark.org/0.25/#example-157) and [159](http://spec.commonmark.org/0.25/#example-159).

kevmoo:
LGTM!

lib/src/block_parser.dart
munificent: 👍

lib/src/block_parser.dart
munificent: Yeah, it does seem pretty complex. However, if I'm reading the spec right, I still think there should be a more straightforward solution. As far as I can tell, parsing it doesn't require complex backtracking. My understanding is that, given a list of lines, you are trying to split it into:

1. The contiguous list of link definitions, each of which may be more than one line.
2. All of the remaining lines.

So, for any given list of lines, the links always come first, and there are no gaps between them. So I think parsing it could be something roughly like:

```dart
/// Parses a ref link starting at [start] in [lines]. Returns the number of
/// lines consumed or 0 if a ref link does not start there.
int parseRefLink(List<String> lines, int start) {
  // Basically the same thing as _parseReflinkDefinition(), but return the
  // number of parsed lines or zero on failure...
}

List<String> _extractReflinkDefinitions(
    BlockParser parser, List<String> lines) {
  var start = 0;
  while (true) {
    var consumed = parseRefLink(lines);
    if (consumed == 0) break;
    start += consumed;
  }

  // The rest of the lines are not links.
  return lines.getRange(start);
}
```

Does that work?

srawlins: I think that is similar to my current solution, because it pushes the backtracking complexity into `parseRefLink`.

If `parseRefLink` is given all of the content from some line to the end of the paragraph, it might look like this:

```markdown
[label]: http://url    // already parsed as a link ref def.
[label]: http://url    // [contents] passed to parseRefLink starts here.
"title"
paragraph text.
```

So `parseRefLink` cannot just regexp on the first line and say "oh, `[label]: http://url` looks like a definition! Because it _should_ parse the next line ("title") as well.

However, it might get a few lines in, and then, nope! No title! Like this:

```markdown
[label]:
http://url
"title
with many
words" oops not really a title.
```

Here, `parseRefLink` would have to try like 1 (could be the start of a ref definition), then add 2 (so far so good, but there might be a title), 3 (oops yeah might be a title), 4 (still might), 5 (oops, nope, weird trailing characters; not a title! backtrack to the place earlier with just the label and url)

srawlins:
(BTW @munificent, if my strife for CommonMark edge case compliance looks like an _endorsement_ for CommonMark edge case design choices, it's not! Sometimes I think John MacFarlane is designing CommonMark with interesting parsing challenges just for the parsing challenges, like the 17 rules dictating [Emphasis and Strong Emphasis](http://spec.commonmark.org/0.26/#emphasis-and-strong-emphasis). He's toying with us!)

PR #112, opened 2016-07-19
Remove unused members, properly implement Node in UnparsedContent

kevmoo:


lib/src/block_parser.dart
```diff
@@ -713,8 +708,6 @@ class OrderedListSyntax extends ListSyntax {
 
 /// Parses paragraphs of regular text.
 class ParagraphSyntax extends BlockSyntax {
-  static final _reflinkDefinitionStart = new RegExp(r'[ ]{0,3}\[');
```
srawlins: Oops! I meant to use this in `lineStartsReflinkDefinition` on line 749 (742 on the right), for efficiency. Would you mind using this there instead of deleting it?

kevmoo: Done!

srawlins:
Love it

PR #117, opened 2016-08-04
Be more clear in stats about a successful test

kevmoo:
strict: exact match including whitespace
loose: matches, except for whitespace

While not "strictly" needed, it'd be nice to be as strict as possible


kevmoo:
Ping, @srawlins ?

srawlins:
LGTM. It's good to see we don't have any errors currently :)

PR #123, opened 2016-08-08
Ensure that 1.18.1 and 1.17.1 are running

kevmoo:
But allow stable and dev to fail – since these might be 1.19+
Hoping to get travis fixed soon


srawlins:
lgtm

PR #125, opened 2016-08-09
Add version information to bin/markdown.dart and example app

### LINKED ISSUE #124: Make it easier to see which version dartdoc/pub uses
Someone pointed out that something that works in https://dart-lang.github.io/markdown/ doesn't work in dartdoc. (In this case, it was creating a list without a leading blank line.)

It'd be nice to be able to specify which version of markdown/dartdoc/pub you're interested in.

If that's not possible, we should at least make it clear (1) which version the interactive editor uses and (2) that tools might be using an old version. We could create a doc here and link to it from the live editor with text like "markdown 0.11.0+1 (details)".

Best of all (from my pov) would be for pub and dartdoc to use (or be configurable to use) the latest version of markdown all the time.

srawlins:
Fixes #124 

Now you can do `bin/markdown.dart --version`, and the example app shows the version as well.

The new `tool/travis.sh` checks:

* dartanalyzer
* dartfmt
* tests
* version match between `lib/src/version.dart` and `pubspec.yaml`.


kevmoo:
LGTM

PR #128, opened 2016-08-10
dartfmt

kevmoo:
I had a travis thing to check this, too. Would be good to have here...


srawlins:
Would you mind adding example/ to https://github.com/dart-lang/markdown/blob/master/tool/travis.sh#L15 with this change?

kevmoo:
Closed with 2840592

git gui snafu

PR #130, opened 2016-08-11
Example: cancel typing when user types

### LINKED ISSUE #127: Example: stop robot typing on click/key down in text box
So a user who wants to hack on it doesn't have to fight the robot

srawlins:
Fixes #127


munificent:
:+1:

PR #132, opened 2016-08-11
New GitHub ExtensionSet; example site supports ExtensionSets!

srawlins:
This is a pretty fun add-on to the site: you can switch between different Extension Sets (none, commonMark, gitHub).

Here's some Markdown that renders differently in all three:

```markdown
Inline <em>HTML</em>.

| Tables | are | great |
|---|---|---|
| But | only in | GitHub |
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
   * `dart bin/markdown.dart --version` now shows the package version number.
   * The playground app now shows the version number.
 * Improve autolink parsing.
-* Added new table syntax: `TableSyntax`. This can be used by passing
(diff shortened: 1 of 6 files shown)


srawlins:
Oh, btw, Dartium does _not_ want to render those glyphs :/ but it works when pub served into Chrome and Firefox.

srawlins:
@kevmoo this is ready for review.

kevmoo:
@srawlins no tests? 😉

kevmoo:
...nevermind. I see it's small.

srawlins:
Thanks Kevin! No tests for the example... yet.

PR #144, opened 2016-10-26
Use more lints – related fixes

kevmoo:


srawlins:
Why some commented? strong is commented?

kevmoo:
User error, @srawlins – PTAL

Issue #153, opened 2017-02-22
TableSyntax allows backslash to escape a pipe


tomyeh:
Example from github:

```
| Name     | Character |
| ---      | ---       |
| Backtick | `         |
| Pipe     | \|        |
```

tomyeh:
Fixed

PR #154, opened 2017-02-24
#153: support escape pipe in table

tomyeh:
A quick fix. It doesn't consider the case: `\\|`, but it shall be good enough. After all, most users will do `\\ |` instead.


tomyeh:
I signed it.

PR #157, opened 2017-03-22
Fix #156: ListSyntax preserves line breaks

### LINKED ISSUE #156: ListSyntax doesn't handle line breaks correctly
Markdown:

```
1. line1
line2
line3
```

Result:

```
<ol><li>line1line2line3</li></ol>\n
```

Rather, it shall be

```
<ol><li>line1\nline2\nline3</li></ol>\n
```

tomyeh:
It can be subjective, but I think it is better to preserve the line breaks like Github does.

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -672,7 +672,7 @@ abstract class ListSyntax extends BlockSyntax {
         }
 
         // Anything else is paragraph continuation text.
-        var continuedLine = childLines.last + parser.current;
+        var continuedLine = childLines.last + '\n' + parser.current;
         childLines
           ..removeLast()
           ..add(continuedLine);
--- a/test/original/unordered_lists.unit
+++ b/test/original/unordered_lists.unit
@@ -83,7 +83,8 @@ two</li><li>three</li></ul>
 ===
 
 <<<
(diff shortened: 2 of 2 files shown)


srawlins:
Does this not affect the common mark stats results? You can check with `tool/common_mark_stats.dart`.

kevmoo:
> Does this not affect the common mark stats results?

Causes 254 to start failing (from loose).

kevmoo:
@tomyeh 

try `dart tool/common_mark_stats.dart --update-files` check for diffs...

tomyeh:
Sorry, I didn't notice it. Yes, it broke 254. No easy fix then -- unless we inject a special token, and replace it back to line feed after it. But, it is dirty. Please forget this pull.

```
var continuedLine = childLines.last + TOKEN + parser.current;
...
for (var item in items) {
  var itemParser = parser.document.getBlockParser(item.lines);
  var children = itemParser.parseLines();
  for (child in children)
    child.text.replaceAll(TOKEN, '\n')

kevmoo:
@tomyeh thanks for trying. If you have ideas of a better way to attack this, we'd love to see a refactor.

There are plenty of other implementations you can look at 😄

PR #164, opened 2017-06-08
Fix #162 by properly putting bq lazy continuations on their own line

### LINKED ISSUE #162: Lazy blockquotes are not handled properly
Input:
```
> a
b
```

Output:
```
<blockquote>
<p>ab</p></blockquote>
```
(Note the absence of whitespace between `a` and `b`)

Expected:
```
<blockquote>
<p>a
b</p></blockquote>
```

The ['spec'](https://daringfireball.net/projects/markdown/syntax#blockquote) says: *"Markdown allows you to be lazy and only put the > before the first line of a hard-wrapped paragraph:"*

By the way the library handles the properly blockquoted input as the expected above:
```
> a
> b
```

srawlins:


srawlins:
Yeah #162 is described in the spec, but not covered by a specific example.

PR #169, opened 2017-06-08
Fix #156, same as #164

### LINKED ISSUE #156: ListSyntax doesn't handle line breaks correctly

srawlins:


kevmoo:
Does *this* change the markdown output?

srawlins:
Not the stats, haha, nope. Possibly a shortcoming of the specs.

PR #174, opened 2017-08-18
Better support for auto links

kevmoo:


kevmoo:
> Too bad the *WithId syntaxes had to go for GFM

Eh, it's no help if GitHub doesn't support it. If we care about GitHub compat, it's good to know upfront.

PR #180, opened 2017-10-25
Lints, etc

kevmoo:


kevmoo:
Now we actually test that our tracked markdown flavors don't change – unless it's intentional

PR #187, opened 2017-12-17
Make the ExtensionSet constructor public

srawlins:


kevmoo:
@srawlins – travis is sad

PR #193, opened 2018-01-02
Overhaul the parsing of emphasis and strong emphasis

srawlins:
CommonMark's emphasis and strong emphasis require complex definitions of
"delimiter runs" as well as 17 parsing rules. This implementation was emulating
these rules poorly with some simple regular expressions.

This work examines possible matches for opening and closing (strong) emphasis in
terms of "delimiter runs" and ability to "open emphasis" and "close emphasis",
in order to dramatically improve the spec-matching for Emphasis.

No specs flip from strict/loose to fail with this work.


srawlins:
Thanks!

PR #194, opened 2018-01-03
Improve tab support, and horizontal rules

srawlins:


lib/src/block_parser.dart
```diff
@@ -597,24 +599,39 @@ abstract class ListSyntax extends BlockSyntax {
       return match != null;
     }
 
+    int expandedTabLength(String input) {
```
kevmoo: Needs to be inline? Could be class static? Top-level?

srawlins: You're right. I like static private.

PR #198, opened 2018-01-05
Allow --extension-set to be passed to bin/markdown

srawlins:


pubspec.yaml
```diff
@@ -5,8 +5,11 @@ description: A library for converting markdown to HTML.
 homepage: https://github.com/dart-lang/markdown
 environment:
   sdk: '>=1.12.0 <2.0.0'
+dependencies:
+  args: '^1.0.0'
```
kevmoo: Ugh...I hate adding args to packages that are mostly used as ...packages

How important is shipping a `bin/markdown.dart`? Do we think folks really want/use it?

srawlins: $ pub global activate markdown
Activated **markdown** 1.1.1
$ pub global run markdown --extension-set GitHubWeb
:cry&#8203;: you don't think they love it?
^D
:cry: you don't think they love it?

srawlins: I totally agree though. Why no `bin_dependencies`??

kevmoo: Uh...there's a feature thing we're pondering – but not until after v2 ships

srawlins: I'll deflect with a flash of Whataboutism! If you depend on the little bitty [tuple](https://pub.dartlang.org/packages/tuple#-readme-tab-), you get all of [quiver](https://github.com/dart-lang/tuple/blob/master/pubspec.yaml). If you have [dart_style](https://github.com/dart-lang/dart_style/blob/master/pubspec.yaml) on disk, you have args on disk.

Don't get me wrong, this kills me a little inside. But I'm not sure what the actual consequences are:

* args is surely already in your pub cache
* if markdown's dependency on args is the only reason it's in your web app, it will get tree shaken out by dart2js
* if markdown's dependency on args is the only reason it's in your VM or flutter app, I think it won't ever load into the VM, as long as it's never imported?
* it _will_ ship in a Flutter app, right? The entire contents of every dependency package is included, including tests, no?

kevmoo:
Hrm...yeah.

We had issues when args moved from v0.x to v1.x – weird things had to be
updated – Like CSS lib.

I feel like markdown is more like csslib than dart_style.

I'll live, though...

kevmoo:
...and note in the changelog folks can 'pub global activate'

srawlins:
Done!

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@
 * Improve support for tab characters, and horizontal rules.
 * The above fixes raise compliance with the CommonMark specs to 90%, and
   compliance with the GFM specs to 87%.
+* Allow the binary script to take a `--extension-set` option.
+
+  A reminder: You can run `bin/markdown.dart` from anywhere via:
+
+  ```shell
+  $ pub global activate markdown
+  $ pub global run markdown
```
kevmoo: Just `markdown` should work if they have their paths set up correctly...

srawlins: Done.

PR #203, opened 2018-01-13
GFM autolink extension

### LINKED ISSUE #201: Github Flavored Markdown extended autolinks unsupported
[The Github Flavored Markdown spec extends CommonMark's autolinks to support links without `<...>` surrounding them.](https://github.github.com/gfm/#autolinks-extension-)

It appears that the current GFM extension set does not include support for this.
image

Raw links were passed on before in #33, but that was due to them [not being a part of the CommonMark spec](http://spec.commonmark.org/0.24/#autolinks).

Would this project be open to a PR to add support for this?

dudeofawesome:
This is a very much a work in progress. 

Currently, I'm trying to figure out why a few of the tests are failing when there are multiple links to match. It looks like the RegExp only matches the first link. I'm also seeing some inconsistency with scheme-less link matching.

The CommonMark Autolink tests that are failing should be failing with GFM Autolink Extensions enabled.

This will fix #201

### DIFF
--- a/lib/src/extension_set.dart
+++ b/lib/src/extension_set.dart
@@ -35,7 +35,7 @@ class ExtensionSet {
     const FencedCodeBlockSyntax(),
     const HeaderWithIdSyntax(),
     const SetextHeaderWithIdSyntax(),
-    const TableSyntax(),
+    const TableSyntax()
   ], [
     new InlineHtmlSyntax(),
     new StrikethroughSyntax(),
@@ -49,7 +49,7 @@ class ExtensionSet {
   /// [GitHub flavored Markdown]: https://github.github.com/gfm/
   static final ExtensionSet gitHubFlavored = new ExtensionSet([
     const FencedCodeBlockSyntax(),
-    const TableSyntax(),
+    const TableSyntax()
   ], [
     new InlineHtmlSyntax(),
     new StrikethroughSyntax(),
--- a/lib/src/extension_set.dart
+++ b/lib/src/extension_set.dart
@@ -35,11 +35,12 @@ class ExtensionSet {
     const FencedCodeBlockSyntax(),
     const HeaderWithIdSyntax(),
     const SetextHeaderWithIdSyntax(),
-    const TableSyntax()
+    const TableSyntax(),
   ], [
     new InlineHtmlSyntax(),
     new StrikethroughSyntax(),
     new EmojiSyntax(),
+    new AutolinkExtensionSyntax(),
   ]);
 
   /// The [gitHubFlavored] extension set is close to compliance with the [GitHub
@@ -48,10 +49,11 @@ class ExtensionSet {
   /// [GitHub flavored Markdown]: https://github.github.com/gfm/
   static final ExtensionSet gitHubFlavored = new ExtensionSet([
     const FencedCodeBlockSyntax(),
-    const TableSyntax()
+    const TableSyntax(),
   ], [
     new InlineHtmlSyntax(),
     new StrikethroughSyntax(),
+    new AutolinkExtensionSyntax(),
   ]);
 
   /// The deprecated name for the [gitHubFlavored] extension set.
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -290,7 +290,7 @@ class AutolinkExtensionSyntax extends InlineSyntax {
 
   @override
   bool tryMatch(InlineParser parser) {
-    var startMatch = pattern.matchAsPrefix(parser.source.substring(parser.pos));
+    var startMatch = pattern.matchAsPrefix(parser.source, (parser.pos > 0 ? parser.pos - 1 : 0));
     if (startMatch != null) {
       // Write any existing plain text up to this point.
       parser.writeText();
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -303,11 +303,9 @@ class AutolinkExtensionSyntax extends InlineSyntax {
 
   @override
   bool onMatch(InlineParser parser, Match match) {
-    print('HELP! I am trapped in a dart factory');
     var url = match[1];
     var href = url;
     var matchLength = url.length;
-    print('url: $url');
 
     if (url.startsWith(new RegExp(r'(\s|\>)'))) {
       url = url.substring(1, url.length - 1);
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -315,19 +315,17 @@ class AutolinkExtensionSyntax extends InlineSyntax {
       matchLength--;
     }
 
-    /** Prevent accidental standard autolink matches */
+    // Prevent accidental standard autolink matches
     if (url.endsWith('>') && parser.source[parser.pos - 1] == '<') {
       return false;
     }
 
-    /**
-     * When an autolink ends in ), we scan the entire autolink for the total
-     * number of parentheses. If there is a greater number of closing
-     * parentheses than opening ones, we don’t consider the last character
-     * part of the autolink, in order to facilitate including an autolink
-     * inside a parenthesis:
-     * https://github.github.com/gfm/#example-600
-     */
+    // When an autolink ends in ), we scan the entire autolink for the total
+    // number of parentheses. If there is a greater number of closing
+    // parentheses than opening ones, we don’t consider the last character
+    // part of the autolink, in order to facilitate including an autolink
+    // inside a parenthesis:
+    // https://github.github.com/gfm/#example-600
     if (url.endsWith(')')) {
       final opening = new RegExp(r'\(').allMatches(url).length;
       final closing = new RegExp(r'\)').allMatches(url).length;
@@ -338,12 +336,10 @@ class AutolinkExtensionSyntax extends InlineSyntax {
       }
     }
 
-    /**
-     * Trailing punctuation (specifically, ?, !, ., ,, :, *, _, and ~) will
-     * not be considered part of the autolink, though they may be included
-     * in the interior of the link:
-     * https://github.github.com/gfm/#example-599
-     */
+    // Trailing punctuation (specifically, ?, !, ., ,, :, *, _, and ~) will
+    // not be considered part of the autolink, though they may be included
+    // in the interior of the link:
+    // https://github.github.com/gfm/#example-599
     final trailingPunc =
         new RegExp('$truncatingPunctuationPos*' + r'$').firstMatch(url);
     if (trailingPunc != null) {
@@ -352,14 +348,12 @@ class AutolinkExtensionSyntax extends InlineSyntax {
       matchLength -= trailingPunc[0].length;
     }
 
-    /**
-     * If an autolink ends in a semicolon (;), we check to see if it appears
-     * to resemble an
-     * [entity reference](https://github.github.com/gfm/#entity-references);
-     * if the preceding text is & followed by one or more alphanumeric
-     * characters. If so, it is excluded from the autolink:
-     * https://github.github.com/gfm/#example-602
-     */
+    // If an autolink ends in a semicolon (;), we check to see if it appears
+    // to resemble an
+    // [entity reference](https://github.github.com/gfm/#entity-references);
+    // if the preceding text is & followed by one or more alphanumeric
+    // characters. If so, it is excluded from the autolink:
+    // https://github.github.com/gfm/#example-602
     if (url.endsWith(';')) {
       final entityRef = new RegExp(r'\&[a-zA-Z0-9]+;$').firstMatch(url);
       if (entityRef != null) {
@@ -370,7 +364,7 @@ class AutolinkExtensionSyntax extends InlineSyntax {
       }
     }
 
-    /** The scheme http will be inserted automatically */
+    // The scheme http will be inserted automatically
     if (!href.startsWith(new RegExp(r'(?:https?|ftp)\:\/\/'))) {
       href = 'http://$href';
     }
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -277,13 +277,24 @@ class AutolinkSyntax extends InlineSyntax {
 /// Matches autolinks like `http://foo.com`.
 class AutolinkExtensionSyntax extends InlineSyntax {
   /// Broken up parts of the autolink regex for reusability and readability
+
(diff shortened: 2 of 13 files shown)


dudeofawesome:
I signed it!

tool/gfm_tests.json
```diff
@@ -5461,7 +5461,7 @@
     "example": 605,
     "start_line": 9025,
     "end_line": 9029,
-    "section": "Autolinks (extension)",
+    "section": "Autolinks Email (extension)",
```
kevmoo: FYI: this is generated from the GFM source. Shouldn't hand edit.

dudeofawesome: Alright, I can definitely revert that. It was just for testing to cut down on the number of tests being run.

kevmoo: Please! I just checked the latest GFM output – these haven't changed

dudeofawesome: I don't think that this is quite ready to merge yet, so I'd like to leave the modified tests as is for now to ease my dev cycle.

I'm still trying to figure out what is causing it to fail some of the Autolink (extension) tests. I'm not seeing why it's failing [example-598](https://github.com/dudeofawesome/dart-lang-markdown/blob/0a503e882d593f6402b215de9e0712e7b40183f8/tool/gfm_tests.json#L5382) and [example-599](https://github.com/dudeofawesome/dart-lang-markdown/blob/0a503e882d593f6402b215de9e0712e7b40183f8/tool/gfm_tests.json#L5393). I also am not sure how to fix [example-604](https://github.com/dudeofawesome/dart-lang-markdown/blob/0a503e882d593f6402b215de9e0712e7b40183f8/tool/gfm_tests.json#L5448).

srawlins:
Woohoo I see only one failed test in Travis!

tool/common_mark_stats.txt
```diff
@@ -24,4 +24,4 @@
   11 of   11 – 100.0%  Tabs
    3 of    3 – 100.0%  Textual content
   19 of   19 – 100.0%  Thematic breaks
- 561 of  624 –  89.9%  TOTAL
+ 558 of  624 –  89.4%  TOTAL
```
kevmoo: ooo...regression here!

tool/gfm_tests.json
kevmoo: Please revert the changes to this file. It's autogenerated...

tool/common_mark_stats.json
```diff
@@ -279,12 +279,12 @@
   "115": "fail"
  },
  "Hard line breaks": {
-  "605": "strict",
```
kevmoo: I think you fixed these, too....looking at travis failures. May want to regenerate

lib/src/extension_set.dart
```diff
@@ -35,11 +35,12 @@ class ExtensionSet {
     const FencedCodeBlockSyntax(),
     const HeaderWithIdSyntax(),
     const SetextHeaderWithIdSyntax(),
-    const TableSyntax()
+    const TableSyntax(),
```
kevmoo: please avoid unrelated changes...just makes the diffs easier to read

lib/src/extension_set.dart
```diff
@@ -48,10 +49,11 @@ class ExtensionSet {
   /// [GitHub flavored Markdown]: https://github.github.com/gfm/
   static final ExtensionSet gitHubFlavored = new ExtensionSet([
     const FencedCodeBlockSyntax(),
-    const TableSyntax()
+    const TableSyntax(),
```
kevmoo: ditto

tool/common_mark_stats.txt
kevmoo: Gotta get rid of these regressions...

dudeofawesome: Whoops, I was only generating the test files for GFM

tool/gfm_stats.json
```diff
@@ -34,21 +34,21 @@
   "589": "strict",
   "590": "loose",
   "591": "loose",
-  "592": "loose",
```
kevmoo: Trading failures is less than ideal...but our total numbers are better.

dudeofawesome: If you look at the test cases for 592 and 595, you'll see that there's an incompatibility between the GFM autolink extensions and the original CM autolinks. 

Example 592 specifies that `< http://foo.bar >` should not get matched as a link, but the GFM autolink extension doesn't specify anything that would prevent that from matching.

Example 595 is literally just a link `http://example.com`, which obviously should be found by GFM autolinks.

dudeofawesome: On that note, I'm authoring an issue right now on the [github/cmark](https://github.com/github/cmark) repo to address this. 

@srawlins You had mentioned offline that there were a few other projects using the cmark-gfm examples as test cases. Can you link me to those?

srawlins: Er, I thought pulldown-cmark (Rust) was... but they appear to be [using regular CommonMark](https://github.com/google/pulldown-cmark/blob/8b211fde18df2b974163716189cd7b197bdbc6b4/src/passes.rs#L100) tests. I'm not sure of others. In any case, I think it would be hard for them to argue that the examples should remain.

srawlins: @kevmoo I'd like to comment somewhere that 2 tests are failing because of a bug in github/cmark's tests, but not sure where to comment on that... maybe we can all just remember in our brains.

kevmoo:
Honestly, since we're careful about diffs in the generated output, I think
just putting the comment in the txt summary file is okay

WDYT?

tool/gfm_stats.json
dudeofawesome: [Here's the issue I opened with github/cmark](https://github.com/github/cmark/issues/81).

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
   }
 }
 
+/// Matches autolinks like `http://foo.com`.
+class AutolinkExtensionSyntax extends InlineSyntax {
+  static const START = r'(?:^|[\s*_~(>])';
```
srawlins: Dart has moved to lowerSnakeCase for const stuff. Could you update these variable names?

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+  static const SCHEME = r'(?:(?:https?|ftp):\/\/|www\.)';
+  static const DOMAIN_PART = r'\w\-';
+  static const DOMAIN = '[$DOMAIN_PART][$DOMAIN_PART.]+';
+  static const PATH = r'[^\s<]*';
+  // static const TRUNCATING_PUNCTUATION_NEG = r'[^\s<\?\!\.\,\:\*\_\~]';
```
srawlins: Delete.

lib/src/inline_parser.dart
srawlins: Could you comment on what the static strings are, just as a group.

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+  static const TRUNCATING_PUNCTUATION_NEG = r'';
```
srawlins: Could you put a specific comment for each of these strings. What is a "truncating punctuation"? You could even just mention to look at the comment for `trailingPunc` below.

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+  static const TRUNCATING_PUNCTUATION_POS = r'[?!.,:*_~]';
+
+  AutolinkExtensionSyntax()
+      : super('$START(($SCHEME)($DOMAIN)($PATH))$TRUNCATING_PUNCTUATION_NEG');
+
+  @override
+  bool tryMatch(InlineParser parser) {
+    var startMatch = pattern.matchAsPrefix(
+        parser.source, (parser.pos > 0 ? parser.pos - 1 : 0));
+    if (startMatch != null) {
```
srawlins: Let's flip this logic, for less nesting:

```dart
if (startMatch == null) return false;

parser.writeText();
...
```

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+    }
+
+    /**
+     * Trailing punctuation (specifically, ?, !, ., ,, :, *, _, and ~) will
+     * not be considered part of the autolink, though they may be included
+     * in the interior of the link:
+     * https://github.github.com/gfm/#example-599
+     */
+    final trailingPunc =
+        new RegExp('$TRUNCATING_PUNCTUATION_POS*' + r'$').firstMatch(url);
```
srawlins: Can you make this RegExp static as well, so that its not compiled every go.

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+    /**
+     * When an autolink ends in ), we scan the entire autolink for the total
+     * number of parentheses. If there is a greater number of closing
+     * parentheses than opening ones, we don’t consider the last character
+     * part of the autolink, in order to facilitate including an autolink
+     * inside a parenthesis:
+     * https://github.github.com/gfm/#example-600
+     */
+    if (url.endsWith(')')) {
+      final opening = new RegExp(r'\(').allMatches(url).length;
```
srawlins: Let's not use RegExp just to count the number of parens. Can you write a private helper method that just walks the url and counts opening and closing? Should be much faster.

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+      if (entityRef != null) {
+        // Strip out HTML entity reference
+        url = url.substring(0, url.length - entityRef[0].length);
+        href = href.substring(0, href.length - entityRef[0].length);
+        matchLength -= entityRef[0].length;
+      }
+    }
+
+    /** The scheme http will be inserted automatically */
+    if (!href.startsWith(new RegExp(r'(?:https?|ftp)\:\/\/'))) {
```
srawlins: Let's just use some strings, instead of a RegExp:

```dart
if (!href.startsWith('http://') &&
    !href.startsWith('https://') &&
    !href.startsWith('ftp://'))
```

lib/src/inline_parser.dart
```diff
@@ -274,6 +274,118 @@ class AutolinkSyntax extends InlineSyntax {
...
+      href = 'http://$href';
+    }
+
+    final anchor = new Element.text('a', escapeHtml(url));
+    anchor.attributes['href'] = Uri.encodeFull(href);
+    parser.addNode(anchor);
+
+    parser.consume(matchLength);
+    return false;
```
srawlins: Why are we returning false here?

lib/src/inline_parser.dart
dudeofawesome: What is the preferred comment style for Dart? I see a lot of `///` elsewhere in the document, but a few `//` as well.

Also, I'll update my other comments to match the preferred style.

lib/src/inline_parser.dart
dudeofawesome: This is actually almost identical to the `super.tryMatch`, which has the logic in the current order. Should we change it in both places?

lib/src/inline_parser.dart
dudeofawesome: Could I switch this to `url.allMatches('(').length;` for a similar speed improvement, or does this still use a RegExp underneath? The documentation doesn't seem to explicitly say that it does.

lib/src/inline_parser.dart
dudeofawesome: `onMatch` returns whether the caller should advance the parser by `match[0].length`. Since I'm calling `parser.consume(matchLength);` above, it needs to not consume the match length again.

dudeofawesome:
@kevmoo `dart tool/stats.dart --update-files` only updates the CM stats files. Is that intentional? If so, is there a way to run both CM and GFM in the same command? That's been throwing me off a few times.

srawlins:
@dudeofawesome you can run `dart tool/stats.dart --update-files --flavor gfm`

lib/src/inline_parser.dart
srawlins: OK sg.

lib/src/inline_parser.dart
srawlins: Ah, I see, unfortunate that you have to override just for that little backtracking offset.

Not the end of the world though. You can leave the super.tryMatch logic alone. I might clean it up later.

tool/gfm_stats.json
srawlins: Thanks that works.

dudeofawesome:
Just did, no changes to commit.

lib/src/inline_parser.dart
dudeofawesome: Perhaps I could add an optional offset parameter to `super.tryMatch`, and then just call that, so that I can remove all of the redundant code here?

srawlins: Yeah that's probably not too unruly. Let's see how it looks.

kevmoo:
Of course I just messed it up by merging in the changes from @srawlins 

Sorry, @dudeofawesome – one more time?

dudeofawesome:
Rebased and updated!

lib/src/inline_parser.dart
```diff
@@ -274,6 +275,128 @@ class AutolinkSyntax extends InlineSyntax {
...
+  static final regExpTrailingPunc =
+      new RegExp('$truncatingPunctuationPositive*' + r'$');
+  static final regExpEndsWithColon = new RegExp(r'\&[a-zA-Z0-9]+;$');
+  static final regExpWhiteSpace = new RegExp(r'\s');
+
+  AutolinkExtensionSyntax() : super('$start(($scheme)($domain)($path))');
+
+  @override
+  bool tryMatch(InlineParser parser, [int startMatchPos]) {
+    return super.tryMatch(parser, parser.pos > 0 ? parser.pos - 1 : 0);
```
srawlins: Yeah this is very nice now.

srawlins:
Code looks good; Travis is happy; GFM compliance up from 88% to **89%** Woohoo  🙌  🎉 🍾  ! Thanks much @dudeofawesome!

PR #204, opened 2018-01-19
Rename the Link class to LinkReference

srawlins:
After receiving a fair amount of feedback on #202, I think it would be good to do my best to open some preliminary PRs to overhaul links. This is the first, renaming `Link` to `LinkReference`, so that I can create a new class, `Link`, that will represent something like a link, or a possible link. The `LinkReference` is a different concept. Other changes (mostly just moving to CommonMark terms):

* `Link.id` is now `LinkReference.label`
* `Link.url` is now `LinkReference.destination`
* `Document.refLinks` is now `Document.linkReferences`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-## 1.1.2-dev
+## 2.0.0-dev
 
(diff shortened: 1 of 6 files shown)


srawlins:
CC @nex3

PR #208, opened 2018-03-23
Make the fields in Document read-only

kevmoo:


srawlins:
Is this a breaking change?

kevmoo:
Strictly speaking, yes!

Although I'm guessing we wouldn't want folks to tweak these...

kevmoo:
@srawlins updated changelog.

Also confirmed that dartdoc and the package site don't use these...

srawlins:
Needs a quick rebase, then LGTM!

PR #218, opened 2018-04-30
Add Document option to prevent HTML escaping.

### LINKED ISSUE #211: md.Document parses the & and < symbols as HTML instead of raw text
This caused https://github.com/flutter/flutter_markdown/issues/42.  Since Document parses Markdown into an AST, it shouldn't be creating HTML-specific text entities.  Doing this makes it impossible to use the parser to create non-HTML representations of Markdown, including Flutter widgets, without hacking in style overrides.

Is it possible to adjust Document to produce a proper AST that we can render to HTML or other representations separately?

westy92:
Fixes #211.

I wasn't sure how to structure the added test since I didn't see any other `Document`-specific tests.

### DIFF
--- a/lib/src/document.dart
+++ b/lib/src/document.dart
@@ -13,7 +13,7 @@ class Document {
   final ExtensionSet extensionSet;
   final Resolver linkResolver;
   final Resolver imageLinkResolver;
-  final bool guardHtml;
+  final bool encodeHtml;
   final _blockSyntaxes = new Set<BlockSyntax>();
   final _inlineSyntaxes = new Set<InlineSyntax>();
 
@@ -26,7 +26,7 @@ class Document {
       ExtensionSet extensionSet,
       this.linkResolver,
       this.imageLinkResolver,
-      this.guardHtml = false})
+      this.encodeHtml = true})
       : this.extensionSet = extensionSet ?? ExtensionSet.commonMark {
     this._blockSyntaxes
       ..addAll(blockSyntaxes ?? [])
--- a/lib/src/document.dart
+++ b/lib/src/document.dart
@@ -13,6 +13,7 @@ class Document {
   final ExtensionSet extensionSet;
   final Resolver linkResolver;
   final Resolver imageLinkResolver;
+  final bool guardHtml;
   final _blockSyntaxes = new Set<BlockSyntax>();
   final _inlineSyntaxes = new Set<InlineSyntax>();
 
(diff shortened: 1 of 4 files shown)


test/document_test.dart
```diff
@@ -0,0 +1,18 @@
...
+
+import 'package:markdown/markdown.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Document', () {
+    test('guardHtml prevents less than and ampersand escaping', () {
+      var document = new Document(guardHtml: true);
+      var result = document.parseInline('< &');
+      expect(result, hasLength(1));
```
westy92: I tried:
```dart
var expected = [new Text('< &')];
expect(result, equals(expected));
```
But I got:
```
  Expected: [Instance of 'Text']
    Actual: [Instance of 'Text']
     Which: was <Instance of 'Text'> instead of <Instance of 'Text'> at location [0]
```

lib/src/document.dart
```diff
@@ -13,6 +13,7 @@ class Document {
   final ExtensionSet extensionSet;
   final Resolver linkResolver;
   final Resolver imageLinkResolver;
+  final bool guardHtml;
```
srawlins: Hmm, `guardHtml` doesn't intuitively tell me what it does, or which way is normal. How about `encodeHtml` which would default to true?

test/document_test.dart
srawlins: Ouch, sorry about that. Want to add a `toString()` to [Text](https://github.com/dart-lang/markdown/blob/master/lib/src/ast.dart#L59)?

westy92: I tried adding this to `Text`:

```dart
String toString() => '<text: $text>';
```

Test:
```dart
test('guardHtml prevents less than and ampersand escaping', () {
  var document = new Document(guardHtml: true);
  var result = document.parseInline('< &');
  var expected = [new Text('< &')];
  expect(result, expected);
});
```

Result:
```
00:24 +1586 -1: test/document_test.dart: Document guardHtml prevents less than and ampersand escaping [E]                                                             
  Expected: [Text:<text: < &>]
    Actual: [Text:<text: < &>]
     Which: was Text:<<text: < &>> instead of Text:<<text: < &>> at location [0]
  
  package:test                  expect
  test/document_test.dart 14:7  main.<fn>.<fn>
  
00:29 +2484 -1: Some tests failed.  
```

westy92:
@srawlins I addressed all your comments. :tada:

srawlins:
Sorry for the delay. Thanks for the contribution, I love it!

PR #222, opened 2018-05-26
Rename app.dart to example.dart

bcko:
rename so that this example can show up on the pub site

 https://www.dartlang.org/tools/pub/package-layout#examples

related issue : dart-lang/site-www#413


srawlins:
@kevmoo will this still work with GitHub pages, and `tool/update-gh-pages.sh`?

kevmoo:
Oh yeah. Don't care about the dart name – just 'index.html'

PR #223, opened 2018-06-22
dart2 fixes

kevmoo:
will land as separate commits


test/blns_test.dart
```diff
@@ -23,7 +23,7 @@ void main() {
   for (var str in blns) {
     test('blns string $index', () {
       var result = markdownToHtml(str);
-      expect(result, new isInstanceOf<String>());
+      expect(result, const TypeMatcher<String>());
```
srawlins: Does the min version of `test` need to be bumped to use this?

kevmoo: Oops! yup

kevmoo: done!

PR #244, opened 2019-04-16
Escape HTML attribute for fenced code blocks, in the info string; 2.0.3

srawlins:


kevmoo:
Looks like you need to re-run the common mark and gfm stats

srawlins:
Bumped.

The spec is [#311](https://spec.commonmark.org/0.28/#example-311). Here's [BabelMark](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%60%60%60+f%26ouml%3B%26ouml%3B%0Afoo%0A%60%60%60) showing that most renderers keep the entity codes as they are, while CommonMark literally changes them to their Unicode character.

I think we should just let this regression in... I'm not sure how to not escape the `&`. From HtmlEscapeMode's [docs](https://api.dartlang.org/stable/2.2.0/dart-convert/HtmlEscapeMode-class.html):

> All modes require escaping of & (ampersand) characters, and may enable escaping of more characters.

Womp womp.

PR #256, opened 2019-07-12
Fix some html entities

srawlins:


srawlins:
Done, thanks for the quick review!

Issue #259, opened 2019-08-23
Render inconsistencies with three backticks (code block) (compared to GitHub)


isoos:
Original report on pub site: https://github.com/dart-lang/pub-dev/issues/2697

Root cause:
```
print(markdownToHtml(' * ``` a ``` b'''));
// prints: 
// <ul><li>
// <pre><code class="language-a"></code></pre></li></ul>
//
// note: the "b" disappearing, and "a" is converted to language annotation
//
// expected:
// <ul><li><code>a</code> b</li></ul>
```

GitHub renders it as the user expected:
https://github.com/DokkarRachidReda/Fancy-Flutter-Alert-Dialog#use-the-package

jonasfj:
I would think expected output is:
```
// <ul><li><pre><code>a</code></pre> b</li></ul>
```
But I could be wrong, we probably need to refer to a spec for markdown, if one exists :)

jonasfj:
See: https://spec.commonmark.org/0.29/#code-spans

> A code span begins with a backtick string and ends with a backtick string of equal length.

It seems `package:markdown` has issues with 3 or more backticks.

srawlins:
Fix pushed in markdown 2.1.0.

PR #260, opened 2019-08-28
Fix parsing of inline code blocks with multiple backticks

srawlins:
Fixes https://github.com/dart-lang/markdown/issues/259


srawlins:
The failed internal check is fine; we have a broken patch file.

lib/src/block_parser.dart
```diff
@@ -404,12 +406,28 @@ class CodeBlockSyntax extends BlockSyntax {
...
 class FencedCodeBlockSyntax extends BlockSyntax {
-  RegExp get pattern => _codePattern;
+  RegExp get pattern => _codeFencePattern;
 
   const FencedCodeBlockSyntax();
 
+  bool canParse(BlockParser parser) {
+    var match = pattern.firstMatch(parser.current);
+    if (match == null) return false;
+    final codeFence = match.group(1);
```
munificent: Use `var` here and below or `final` above.

lib/src/block_parser.dart
```diff
@@ -404,12 +406,28 @@ class CodeBlockSyntax extends BlockSyntax {
...
+    final infoString = match.group(2);
+    // From the CommonMark spec:
+    //
+    // > If the info string comes after a backtick fence, it may not contain
+    // > any backtick characters.
+    if (codeFence.codeUnitAt(0) == $backquote &&
```
munificent: How about:

```dart
return codeFence.codeUnitAt(0) != $backquote ||
    !infoString.codeUnits.contains($backquote));
```

kevmoo:
Land it!

PR #262, opened 2019-09-23
Fix escaping in an indented code block

### LINKED ISSUE #261: encodeHtml does not get considered inside any BlockSynatx
`encodeHtml` is not considered for example for inside the `CodeBlockSyntax` leading to any `<` being escaped inside of `flutter_markdown` which displays it as normal.

For example here https://github.com/dart-lang/markdown/blob/master/lib/src/block_parser.dart#L401 it is escaped nonetheless of this flag.
Screenshot 2019-09-20 at 16 08 22

srawlins:
Fixes #261


srawlins:
Yes, thanks! Done.

PR #266, opened 2019-09-27
Enforce more strict analysis

srawlins:


kevmoo:
@srawlins – travis is sad. 😢

Issue #271, opened 2019-11-01
parsing large markdown is slow on Flutter


jerrywell:
parsing the following large markdown (roughly 65536 words) is slow. (version 2.1.1)

```
markdown version: 2.1.1
Flutter version: 1.10.14 (latest tag version)
```

the measurement on Samsung's S7:

with JIT: ~800ms
with AOT: ~1500ms

```dart
  final markdown = '''
[test](www.google.com)

~~test~~
zzzz

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla non augue nec sem fermentum pharetra vitae vel risus. Etiam metus sapien, condimentum vestibulum porta at, feugiat non dolor. Etiam quis nisl at velit tincidunt tempor. Nullam nec euismod lectus. Fusce non libero a est blandit commodo. Duis augue libero, congue non tristique eget, porttitor quis eros. Praesent a tellus massa.

Interdum et malesuada fames ac ante ipsum primis in faucibus. Sed sit amet risus non quam pulvinar semper. Nunc et venenatis odio, et condimentum metus. Nulla lorem libero, consequat pellentesque imperdiet eget, hendrerit a nulla. Mauris aliquam orci ac augue mollis, et malesuada massa mollis. Vestibulum auctor accumsan suscipit. Morbi quis risus nec elit ornare accumsan quis id est. Ut odio quam, tincidunt id consectetur a, molestie ut tortor. Etiam ac diam vitae eros ornare aliquam non non ante. Nullam eget commodo nibh, in eleifend nisl. Mauris dapibus bibendum nisl, non pretium dolor tristique in. Nullam convallis metus ut nunc elementum, ut mollis ligula aliquet. Mauris egestas, turpis in imperdiet fringilla, nunc sem ultrices libero, nec gravida metus nisl ac sapien. Etiam non lectus sit amet augue maximus commodo sit amet viverra lectus. Aliquam bibendum magna nec felis euismod commodo.

[...]

PR #274, opened 2019-11-14
Reduce time to parse a large Markdown document by about half

srawlins:
Largely addresses #271 but I might squeeze some more performance, so not closing yet.

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -136,13 +136,13 @@ class BlockParser {
   /// Gets whether or not the current line matches the given pattern.
   bool matches(RegExp regex) {
     if (isDone) return false;
-    return regex.firstMatch(current) != null;
+    return regex.hasMatch(current);
   }
 
   /// Gets whether or not the next line matches the given pattern.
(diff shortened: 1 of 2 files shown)


srawlins:
@jonahwilliams, I don't think you can officially review this, but want to take a look and give an LGTM if you like?

lib/src/block_parser.dart
```diff
@@ -136,13 +136,13 @@ class BlockParser {
   /// Gets whether or not the current line matches the given pattern.
   bool matches(RegExp regex) {
     if (isDone) return false;
-    return regex.firstMatch(current) != null;
+    return regex.hasMatch(current);
```
jonahwilliams: Do you know why this is so much faster?

srawlins: Those might not affect performance, but they're simpler, and they _might_ affect performance, if it means no Match object allocations.

jonahwilliams: Ahh, okay - the performance gain here is coming from the smaller regexp check you added

jonahwilliams:
LGTM

PR #288, opened 2020-05-26
Overhaul table row parsing

### LINKED ISSUE #287: wrong data in table
## markdown source

```
| header 1 | header 2 |
| --- | --- |
| a == 3 \|\| b > 4 | row 1 col 2 |
```

## parse code

```
    final content = mdFile.readAsStringSync();
    final m.Document document = m.Document(
        extensionSet: m.ExtensionSet.gitHubFlavored,
        encodeHtml: false);
    final List<String> lines = content.split(RegExp(r'\r?\n'));
    final nodes = document.parseLines(lines);
```
then I log all `nodes`, I got it

## parsed data logs

```
Tag:table   attr:{}
  Tag:thead   attr:{}
    Tag:tr   attr:{}
      Tag:th   attr:{}
        Text:header 1
      Tag:th   attr:{}
        Text:header 2
  Tag:tbody   attr:{}
    Tag:tr   attr:{}
      Tag:td   attr:{}
        Text:a == 3 ||b > 4
      Tag:td   attr:{}
        Text:row 1 col 2
```
there should be a `space`

image

but the result seems to be wrong

image

Is there a bug? Or I'm using it in a wrong way?🤔️

srawlins:
Overhaul table row parsing in order to fix #287 

The current table row parsing is very regular-expression heavy, and not as efficient as just character-walking. So that's the general idea of this PR:

* Fix parsing of a table row like

  ```markdown
  | abc \| def |
  ```

  so that the whitespace after the escaped pipe is maintained.

* Remove some regular expressions; while these seem innocent and simple, they've been known to be very slow to evaluate on Flutter.

* Bump minimum SDK version, in order to support UI-as-code. 2.3.0 through 2.5.0 have a bug with strict-raw-types, so I'm bumping to 2.6.0, released last November.


srawlins:
Not sure why pedantic is being more restrictive on this PR... may be fixed by https://github.com/dart-lang/markdown/pull/289.

srawlins:
> I don't know whether calling codeUnitAt() repeatedly or converting line to a codeUnits list and indexing on it is faster

! This is a really good question. I've never looked into it. But this has the real possibility of changing performance...

asjqkkkk:
Hi, When will this PR be merged into the latest version?
I'm waiting for the update😂

srawlins:
Apologies for the delay. Shipped in 2.1.5.

PR #295, opened 2020-07-08
Remove redundant LinkSyntax; remove erroneous comment

alechenninger:
The LinkSyntax is added, per the comment within _defaultSyntaxes, in the
constructor, so I believe it was not intended to be in the default
syntaxes list without the right resolver set.

The comment within _htmlSyntaxes didn't make sense to me, so I believe
it to be an error (copy and paste?).

Both changes are no-ops, just clean ups I noticed while looking through
the code. Hopefully I have not misunderstood.

Thanks!


srawlins:
Thanks much!!

PR #296, opened 2020-07-14
Small modifications for null-safety

srawlins:
These include nullability hints for the migration tool, local variables for
non-promotable fields, and local variables for regex matches.

### DIFF
--- a/tool/stats_lib.dart
+++ b/tool/stats_lib.dart
@@ -74,10 +74,10 @@ class CommonMarkTestCase {
   factory CommonMarkTestCase.fromJson(Map<String, dynamic> json) {
     return CommonMarkTestCase(
         json['example'] as int,
-        json['section'] as String,
+        json['section'] as String /*!*/,
         json['start_line'] as int,
         json['end_line'] as int,
-        json['markdown'] as String,
+        json['markdown'] as String /*!*/,
         json['html'] as String);
   }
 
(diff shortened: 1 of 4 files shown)


tool/stats_lib.dart
```diff
@@ -74,10 +74,10 @@ class CommonMarkTestCase {
   factory CommonMarkTestCase.fromJson(Map<String, dynamic> json) {
     return CommonMarkTestCase(
         json['example'] as int,
-        json['section'] as String,
+        json['section'] as String /*!*/,
```
kevmoo, line 77: This (and below) seem really weird. `as String` should completely imply "not null". `!` shouldn't me needed!

kevmoo: I guess when migration is done these will be eliminated. This is just to hint the tool?

kevmoo:
Are you going to create a `null-safety` branch here soonish?

tool/stats_lib.dart
srawlins: Yeah, the migration tool sometimes decides that that should be a nullable string.

srawlins:
Haha I don't desire to. markdown only depends on args and charcode though, so I certainly could. It might be after 2.9.0 final is released though, which would be nice, as I could avoid the branch then.

kevmoo:
> Haha I don't desire to. markdown only depends on args and charcode though, so I certainly could. It might be after 2.9.0 final is released though, which would be nice, as I could avoid the branch then.

`args` is going to be "fun" 😄

PR #304, opened 2020-08-14
Make pre-define extension sets immutable

mjordan56:
This change makes the pre-defined extension sets, none, commonMark, gitHubFlavored, and gitHubWeb, immutable. This is a breaking change only if consumers of the markdown package are modifying the blockSyntaxes or inlineSyntaxes lists of a pre-defined extension set. Causing this type of use of the pre-defined extension sets to break is probably a good thing since modified pre-defined extension sets could have unexpected or unintended side-effects.

I tested this change against the blns_test, document_test, markdown_test, and version_test. All tests passed without errors.


srawlins:
I love it. This will be part of a major version bump, like 3.0.0. I'm going to wait to merge until I create one with some deprecation messages. Hopefully just later today.

PR #308, opened 2020-08-14
Fixed inaccuracies in Extension sets section

mjordan56:
This corrects the README inaccuracies in the Extension sets section. More details could be added to this section but this PR addresses some of the omissions and inaccuracies and is an improvement over the existing information.


srawlins:
Thanks!

PR #311, opened 2020-08-16
Fixes for Issue 310 - Follow Dart file conventions

mjordan56:
Added a longer description of the markdown package to pubspec.yaml and fixed the links in the README.md and CHANGELOG.md files. This should increase the pub points for the markdown package by 20 points from 80/110 to 100/110.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,7 +1,8 @@
 name: markdown
 version: 2.1.8
 
-description: A library for converting markdown to HTML.
+description: A portable Markdown library written in Dart that can parse
+ Markdown into HTML on both the client and server.
 homepage: https://github.com/dart-lang/markdown
 
 executables:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -2,7 +2,7 @@ name: markdown
 version: 2.1.8
 
 description: A portable Markdown library written in Dart that can parse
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -1,7 +1,8 @@
 name: markdown
 version: 2.1.8
 
-description: A library for converting markdown to HTML.
+description: A portable Markdown library written in Dart that can parse
+ Markdown into HTML on both the client and server.
```
srawlins, line 5: I'm not crazy about "on both the client and server," as I don't think those are defined well... are these words required to get the letter count up?

mjordan56: @srawlins No, they aren't needed. To pass the package description length requirement the description needs to be between 60 and 180 characters. If the text "on both the client and server" is removed the description will be 79 characters in length which will meet the requirement. I'll modify the description text and submit the change.

BTW, the expanded description I submitted was taken from the first line of the README.md file:
> A portable Markdown library written in Dart. It can parse Markdown into HTML on both the client
> and server.

If this statement is incorrect or misleading the README.md should probably be updated at some point.

srawlins: Haha, that's interesting. I suspect not written by me. In any case, I love the change.

PR #320, opened 2020-09-11
Improve canEndBlock logic for ListSyntax subclasses

srawlins:
I forgot I was about to release a major version with breaking changes, haha. So now is a good time to change the `canEndBlock` API.


kevmoo (approved):
this seems a lot better!

kevmoo (approved):

PR #346, opened 2020-12-12
Remove dependency on expected_output; fold impl into tool/

srawlins:
This makes it simpler to migrate to null safety. Markdown is more-or-less the only dependant on expected_output.

I also changed the impl to not rely on mirrors. 🙌


kevmoo (approved):

kevmoo:
...once travis is green 😄

PR #347, opened 2020-12-14
Null safe API

srawlins:
Migrate API to null safety.

This does not include some files in bin/, test/, and tool/:

* bin/markdown.dart requires package:args, which has not published a null safe release.
* tool/stats.dart requires package:args.
* tool/stats_lib.dart requires package:html, which has not published a null safe release.
* tool/dartdoc-compare.dart requires package:args.
* test/util.dart requires package:io, which has not published a null safe release.
* test/markdown_test.dart, test/vesion.dart import test/util.dart.

Most APIs upgraded to null safety very smoothly. Very clean change.

### DIFF
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:charcode/charcode.dart';
-import 'package:meta/meta.dart';
 
 import 'ast.dart';
 import 'document.dart';
@@ -47,7 +46,7 @@ class InlineParser {
   ]);
 
   /// The string of Markdown being parsed.
-  final String source;
+  final String? source;
 
   /// The Markdown document this parser is parsing.
   final Document document;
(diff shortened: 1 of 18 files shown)


lib/src/inline_parser.dart
```diff
@@ -346,38 +345,38 @@ abstract class InlineSyntax {
...
 
-    final startMatch = pattern.matchAsPrefix(parser.source, startMatchPos);
+    final startMatch = pattern.matchAsPrefix(parser.source!, startMatchPos);
     if (startMatch == null) return false;
 
     // Write any existing plain text up to this point.
     parser.writeText();
 
-    if (onMatch(parser, startMatch)) parser.consume(startMatch[0].length);
+    if (onMatch(parser, startMatch)) parser.consume(startMatch.match.length);
```
kevmoo, line 379: where does `match` come from?

kevmoo (approved):
Thanks for adding the copyright headers!

lib/src/inline_parser.dart
srawlins: It's a secret!

No, its an extension I dropped into `lib/util.dart`. I think Lasse said it's being considered for the Match class, because `match[0]` should always be non-nullable, but `match[1]` has to be nullable. Lots of `[1]!` and `[2]!` sprinkled :(

PR #348, opened 2020-12-15
Fix leading whitespace after hard break

### LINKED ISSUE #322: Leading Spaces At The Beginning Of Next Line After Hard Break Are Not Ignored
The markdown package incorrectly preserves the leading spaces at the beginning of the next line after a hard break.

Both the [CommonMark](https://spec.commonmark.org/0.29/#example-633) and [GitHub Flavored Markdown](https://github.github.com/gfm/#example-657) specifications note in the section on **Hard line breaks** that "_Leading spaces at the beginning of the next line are ignored_"  (click on the document links to go to the examples in the specifications). The example from both specs show the example Markdown input data:

```
foo・・
・・・・・bar

foo\
・・・・・bar
 ```
should both produce the following HTML output:
```
<p>foo<br />
bar</p>
```
However, the markdown package does not properly trim the trailing spaces. The following sample code:
```
import 'package:markdown/markdown.dart';

void main() {
  print(markdownToHtml('foo  \n     bar'));
}
```
Produces the following incorrect HTML output:
```
<p>foo<br />
     bar</p>
```

srawlins:
Fixes #322 

Also make non-nullable some accidentally nullable APIs.


kevmoo (approved):
You know what makes me happy!

PR #352, opened 2021-01-06
Bump to 4.0.0-nullsafety.0

### LINKED ISSUE #351: publish null-safe package version
It looks like the API is migrated or under-way.

@srawlins: thoughts on publishing?

We'll want this for the linter (https://github.com/dart-lang/linter/issues/2399).

Thanks!

srawlins:
A few dev_dependencies are bumped to their nullsafety versions.

args cannot be bumped yet because test_core does not support the nullsafety version yet.

Fixes #351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.0.0-dev
+## 4.0.0-nullsafety.0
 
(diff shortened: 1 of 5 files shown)


pq (approved):
💎

parlough (approved):

PR #362, opened 2021-03-16
allow latest pkg:io, clean SDk constraint, update pubspec URL

kevmoo:
Use stable setup-dart action, test on oldest supported SDk


srawlins (approved):
SG. Thanks

PR #388, opened 2022-01-20
fix: auto link syntax regex fix

hongfeiyang:
fixed auto link syntax regex error


srawlins:
Thanks for the PR!

Does this fix a certain bug? Is there a test you can add which fails before but passes with this fix?

hongfeiyang:
> Thanks for the PR!
> 
> Does this fix a certain bug? Is there a test you can add which fails before but passes with this fix?

some regex engine like regex101 does not like `/` without a preceding `\`, this is just to make these engines happy :)

PR #394, opened 2022-03-08
Check parser.isDone when title is null in _parseInlineBracketedLink

chenzhiguang:
It will throw an exception when it is an inline bracketed link and there is a backslash before the ending quote.

```dart
markdownToHtml(r'[url](<https://www> "title\")');
```

Screenshot 2022-03-08 at 21 10 56


srawlins (dismissed):
Thanks for the PR! Can you add the test case that causes the crash you showed.

Sorry for the delay.

kevmoo (approved):

PR #395, opened 2022-03-15
Use `Uri.toFilePath()` instead of `Uri.path` for locating tests

telenieko:
When running the test suite on Windows all tests will fail with:

      Failed to load "test\markdown_test.dart": Directory listing failed, path = '/C:/Users/xxxxxxx/markdown\test\original\*' (OS Error: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos., errno = 123)

The important bit to notice is the path: `/C:/`..., the initial forward slash on the path results in an invalid path for Windows.

This can be tracked down to the two calls to `Uri.path`. It would be preferable to use `Uri.toFilePath()` which does take into account  Window's path formats.

You can see the different behaviour on [this dartpad](https://dartpad.dev/f5fa8437fc8fbf7454ef052527308deb).

This PR simply does the switch from `Uri.path` to `Uri.toFilePath()` resulting in tests working on windows.

### DIFF
--- a/test/util.dart
+++ b/test/util.dart
@@ -27,7 +27,7 @@ Future<void> testDirectory(String name, {ExtensionSet? extensionSet}) async {
 Future<String> get markdownPackageRoot async =>
     p.dirname(p.dirname((await Isolate.resolvePackageUri(
             Uri.parse('package:markdown/markdown.dart')))!
-        .path));
+        .toFilePath()));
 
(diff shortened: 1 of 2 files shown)


telenieko:
Closing this PR, I based off my `master` instead of a custom branch.
Submitted again as #396

PR #399, opened 2022-03-21
Remove unnecessary non null assertion in _resolveReferenceLink()

chenzhiguang:
Change getChildren parameter of _resolveReferenceLink to required and remove unnecessary non null assertion.


srawlins (approved):
Thanks!

Issue #367, opened 2021-04-25
Allow providing a parameter to ignore case sensitivity in RegExp when extending InlineSyntax


NamanShergill:

srawlins:
It sounds like you are looking for an extra parameter on the InlineSyntax constructor? That sounds perfectly reasonable.

NamanShergill:
> It sounds like you are looking for an extra parameter on the InlineSyntax constructor? That sounds perfectly reasonable.

Yep! RegExp can have a case sensitive and a couple other parameters passed to it which cannot be passed to the InlineSyntax constructor currently.

PR #400, opened 2022-03-25
Add caseSensitive parameter on the InlineSyntax constructor

chenzhiguang:
Fix: https://github.com/dart-lang/markdown/issues/367

### DIFF
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -356,9 +356,9 @@ abstract class InlineSyntax {
   ///
   /// If [caseSensitive] is disabled, then case is ignored when matching
   /// the [pattern].
-  InlineSyntax(String pattern, {int? startCharacter, bool? caseSensitive})
-      : pattern = RegExp(pattern,
-            multiLine: true, caseSensitive: caseSensitive ?? true),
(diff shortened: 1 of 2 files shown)


lib/src/inline_parser.dart
```diff
@@ -353,8 +353,12 @@ abstract class InlineSyntax {
   ///
   /// If [startCharacter] is passed, it is used as a pre-matching check which
   /// is faster than matching against [pattern].
-  InlineSyntax(String pattern, {int? startCharacter})
-      : pattern = RegExp(pattern, multiLine: true),
+  ///
+  /// If [caseSensitive] is disabled, then case is ignored when matching
+  /// the [pattern].
+  InlineSyntax(String pattern, {int? startCharacter, bool? caseSensitive})
```
srawlins, line 359: Let's make this a non-nullable bool, defaulting to true.

srawlins (changes requested):
Can you add a few tests which might indicate how this would be used? Thanks for the help!

chenzhiguang:
> Can you add a few tests which might indicate how this would be used? Thanks for the help!

Sorry, just saw this comment, I will add some tests later soon.

srawlins (approved):
Thanks!

Issue #335, opened 2020-10-07
Reference Link Labels Cannot Contain Brackets


mjordan56:
The markdown package incorrectly parses reference link labels containing brackets.

Both the [CommonMark](https://spec.commonmark.org/0.29/#example-542) and [GitHub Flavored Markdown](https://github.github.com/gfm/#example-554) specifications note that "_Link labels cannot contain brackets, unless they are backslash-escaped_"  (click on the document links to go to the examples in the specifications). The example from both specs show the example Markdown input data:

```
[foo][ref[]

[ref[]: /uri
 ```
should produce the following HTML output:
```
<p>[foo][ref[]</p>
<p>[ref[]: /uri</p>
```
However, the markdown package does not properly parse this reference link label. The following sample code:
```
import 'package:markdown/markdown.dart';

void main() {
  print(markdownToHtml('[foo][ref[]\n\n[ref[]: /uri'));
}
```
Produces the following incorrect HTML output:
```
<p><a href="/uri">foo</a></p>
```

PR #402, opened 2022-03-25
Do not allow reference link labels contain left brackets

chenzhiguang:
Fix: https://github.com/dart-lang/markdown/issues/335

### DIFF
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -1118,6 +1118,8 @@ class LinkSyntax extends TagSyntax {
           buffer.writeCharCode(char);
         }
         buffer.writeCharCode(next);
+      } else if (char == $lbracket) {
+        return null;
       } else if (char == $rbracket) {
         break;
       } else {
--- a/test/common_mark/links.unit
+++ b/test/common_mark/links.unit
@@ -321,7 +321,7 @@ bar>)</p>
 
 [ref[]: /uri
 <<<
-<p><a href="/uri">foo</a></p>
+<p>[foo][ref[]</p>
 >>> Links - 546
 [foo][ref[bar]]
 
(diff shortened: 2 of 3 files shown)


srawlins:
Does this fix a Commonmark spec then? Can you run `tool/stats.dart --update-files` and `tool/stats.dart --update-files --flavor=gfm` to update the compliance stat files?

chenzhiguang:
> Does this fix a Commonmark spec then? Can you run `tool/stats.dart --update-files` and `tool/stats.dart --update-files --flavor=gfm` to update the compliance stat files?

I have updated the stat files by running `stats.dart`

srawlins:
We should not change the unit files case-by-case, they are meant to be an exact set of test cases from a single version of the commonmark specs. I'll look into bumping them.

srawlins:
Sorry, I totally mis-remembered how we maintain the spec tests. @kevmoo can maybe confirm:

`tool/stats.dart --update-files` will use `tool/common_mark_tests.json` to run specs, and update all of the unit files in `test/`.

Then the unit files represent what the markdown package currently outputs, so that a clean checkout will always have 100% tests passing.

`tool/common_mark_tests.json` is updated separately (download?), e.g. in [#383](https://github.com/dart-lang/markdown/pull/383/files#diff-ad9d1f3e38cc0928c5c6881aaf8edc900b1abadd673e56e6f9a37033add31b1e).

When someone improves compliance (like this PR), the unit file changes, and `tool/*_stats.json` and `tool/*_stats.txt` are updated (via `tool/stats.dart --update-files`).

kevmoo:
I'd have to re-learn, too – even though I wrote it! That sounds right, @srawlins

kevmoo:
So this change seems to make things less broken – but it's still broken.

```
<p>[foo][ref[]</p>
<p>[ref[]: /uri</p>
```

According to spec, this is what we *should* see.

srawlins (approved):
I'll take the PR because it is strictly better than what we have today 😁. To pass commonmark 545, we just need to also change parsing of the link reference (in addition to the reference link, which you solved).

Thanks!

Issue #393, opened 2022-03-02
The ability to disable standard block/inline syntaxes


lotusgate:
### Problem
I need to parse markdown that doesn't comply with the popular specs. I can manage implementing custom syntaxes, but a roadblock I've hit is that the standard block/inline syntaxes, which are forcefully included, parse improperly some data (improperly, as to the custom spec that I have to follow). 

### Example
`<center><a>**some text**</a></center>` is parsed as \*\*some text\*\*, while I want **some text**. The block parser wraps it in a `Text` element, but `UnparsedContent` is what I need.

### Possible Solution
Maybe the `markdownToHtml` function can have two optional parameters - `withStandardBlockSyntaxes` and `withStandardInlineSyntaxes` - that would be `true` by default. Disabling them will prevent the `BlockParser` and `InlineParser` from auto-including the standard syntaxes.

---
Another thing is that I don't need the functionality of all the standard syntaxes and excluding them may improve performance a little.

PR #404, opened 2022-03-28
Add options to disable default syntaxes

chenzhiguang:
Fixes: https://github.com/dart-lang/markdown/issues/393

## Example

```dart
const text = '<center><a>**some text**</a></center>';

final result = markdownToHtml(
  text,
  withDefaultInlineSyntaxes: false,
  withDefaultBlockSyntaxes: false,
  encodeHtml: false,
  inlineSyntaxes: [
    TagSyntax(r'\*+', requiresDelimiterRun: true),
  ],
);

print(result);
```

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -1233,7 +1233,7 @@ class ParagraphSyntax extends BlockSyntax {
   }
 }
 
-/// Walks the parser forward through the lines do not match any [BlockSyntax].
+/// Walks the parser forward through the lines does not match any [BlockSyntax].
 ///
 /// Returns a [UnparsedContent] with the unmatched lines as `textContent`.
 class DummyBlockSyntax extends BlockSyntax {
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -101,7 +101,12 @@ class BlockParser {
 
   BlockParser(this.lines, this.document) {
     blockSyntaxes.addAll(document.blockSyntaxes);
-    blockSyntaxes.addAll(standardBlockSyntaxes);
+
+    if (document.withDefaultBlockSyntaxes) {
+      blockSyntaxes.addAll(standardBlockSyntaxes);
+    } else {
+      blockSyntaxes.add(const DummyBlockSyntax());
+    }
   }
 
   /// Gets the current line.
@@ -1225,5 +1230,33 @@ class ParagraphSyntax extends BlockSyntax {
     parser.document.linkReferences
         .putIfAbsent(label, () => LinkReference(label, destination, title));
     return true;
+  }
+}
+
+/// Walks the parser forward through the lines do not match any [BlockSyntax].
+///
+/// Returns a [UnparsedContent] with the unmatched lines as `textContent`.
+class DummyBlockSyntax extends BlockSyntax {
+  const DummyBlockSyntax();
+
+  @override
+  RegExp get pattern => _dummyPattern;
+
+  @override
+  bool canEndBlock(BlockParser parser) => false;
+
+  @override
+  bool canParse(BlockParser parser) => true;
+
+  @override
+  Node parse(BlockParser parser) {
+    final childLines = <String>[];
+
+    while (!BlockSyntax.isAtBlockEnd(parser)) {
+      childLines.add(parser.current);
+      parser.advance();
+    }
+
+    return UnparsedContent(childLines.join('\n'));
   }
 }
--- a/lib/src/document.dart
+++ b/lib/src/document.dart
@@ -14,10 +14,10 @@ class Document {
   final Resolver? imageLinkResolver;
   final bool encodeHtml;
 
-  /// Whether to have default block syntaxes.
+  /// Whether to use default block syntaxes.
   final bool withDefaultBlockSyntaxes;
 
-  /// Whether to have default inline syntaxes.
+  /// Whether to use default inline syntaxes.
   ///
   /// Need to set both [withDefaultInlineSyntaxes] and [encodeHtml] to
   /// `false` to disable all inline syntaxes including html encoding syntaxes.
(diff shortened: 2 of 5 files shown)


lib/src/block_parser.dart
```diff
@@ -1227,3 +1232,31 @@ class ParagraphSyntax extends BlockSyntax {
     return true;
   }
 }
+
+/// Walks the parser forward through the lines do not match any [BlockSyntax].
```
srawlins, line 1236: grammar nit, something like "... through the lines; does not ..."

lib/src/document.dart
```diff
@@ -10,12 +10,22 @@ import 'inline_parser.dart';
 /// Maintains the context needed to parse a Markdown document.
 class Document {
   final Map<String, LinkReference> linkReferences = <String, LinkReference>{};
-  final ExtensionSet extensionSet;
   final Resolver? linkResolver;
   final Resolver? imageLinkResolver;
   final bool encodeHtml;
+
+  /// Whether to have default block syntaxes.
```
srawlins, line 17: I think "use" would be better than "have" here; "Whether to use ..."

lib/src/document.dart
```diff
@@ -10,12 +10,22 @@ import 'inline_parser.dart';
...
+  final bool withDefaultBlockSyntaxes;
+
+  /// Whether to have default inline syntaxes.
```
srawlins, line 20: I think "use" would be better than "have" here; "Whether to use ..."

lib/src/document.dart
```diff
@@ -28,13 +38,31 @@ class Document {
...
-    _blockSyntaxes
-      ..addAll(blockSyntaxes ?? [])
-      ..addAll(this.extensionSet.blockSyntaxes);
-    _inlineSyntaxes
-      ..addAll(inlineSyntaxes ?? [])
-      ..addAll(this.extensionSet.inlineSyntaxes);
+    this.withDefaultBlockSyntaxes = true,
+    this.withDefaultInlineSyntaxes = true,
+  }) {
+    hasCustomInlineSyntaxes = inlineSyntaxes?.isNotEmpty == true ||
```
srawlins, line 44: Following [Effective Dart](https://dart.dev/guides/language/effective-dart/usage#prefer-using--to-convert-null-to-a-boolean-value), let's use `?? false` here, instead of `== true`.

srawlins, line 44: We should be able to set this in an initializer:

```dart
    this.withDefaultInlineSyntaxes = true,
  }) :
      hasCustomInlineSyntaxes = ...
  {
```

Then the field can be `final` without `late`.

lib/src/document.dart
```diff
@@ -28,13 +38,31 @@ class Document {
...
+        extensionSet?.inlineSyntaxes.isNotEmpty == true;
+
+    _blockSyntaxes.addAll(blockSyntaxes ?? []);
+    _inlineSyntaxes.addAll(inlineSyntaxes ?? []);
+
+    if (extensionSet == null) {
+      // Add blockSyntaxes of ExtensionSet.commonMark by default if
```
srawlins, line 51: I don't think this comment is necessary; it's a rewording of the two lines of code below it.

lib/src/document.dart
```diff
@@ -28,13 +38,31 @@ class Document {
...
+      // withDefaultBlockSyntaxes is true
+      if (withDefaultBlockSyntaxes) {
+        _blockSyntaxes.addAll(ExtensionSet.commonMark.blockSyntaxes);
+      }
+
+      // Add inlineSyntaxes of ExtensionSet.commonMark by default if
```
srawlins, line 57: I don't think this comment is necessary either.

lib/src/html_renderer.dart
```diff
@@ -19,13 +19,19 @@ String markdownToHtml(
   Resolver? linkResolver,
   Resolver? imageLinkResolver,
   bool inlineOnly = false,
+  bool? encodeHtml,
```
srawlins, line 22: Can we use default values here, `bool encodeHtml = true`, to avoid the nullable bool?

srawlins (changes requested):

lib/src/html_renderer.dart
chenzhiguang: Agreed. Fixed

lib/src/document.dart
chenzhiguang: Yep, I have removed it.

lib/src/document.dart
chenzhiguang: Fixed

chenzhiguang: Fixed. This is a good one. Thanks.

lib/src/document.dart
chenzhiguang: Fixed

lib/src/document.dart
chenzhiguang: Fixed

lib/src/block_parser.dart
chenzhiguang: Fixed

srawlins (changes requested):
This looks great! We need some tests. Maybe something simple that:

* no syntaxes; what does that look like?
* no inline syntaxes
* no block syntaxes

test/html_renderer_test.dart
```diff
@@ -0,0 +1,80 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:markdown/markdown.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('markdownToHtml', () {
+    const text = '# Hello **Markdown<em>!</em>**\n***';
+
+    test('with no syntaxes', () {
```
srawlins, line 12: Excellent

srawlins (approved):
Great tests, thanks!

Issue #357, opened 2021-02-19
Incorrect Parsing of Strikethrough Tags When Mixed with Emphasis and Order Changes


mjordan56:
The markdown package incorrectly parses Markdown strikethrough tags that are combined with other emphasis tags and the order of the tags changes. This seems to only happen when the inline Markdown tags are in the same text block.

For example, the following Markdown text:

`**~~first~~** ~~**second**~~`

Should produce:

**~~first~~** ~~**second**~~

But the actual output is:

**~~first~~** \~\~**second**\~\~

Using the following code test code:

```dart
void main() {
  print(markdownToHtml(r'**~~first~~** ~~**second**~~',
      extensionSet: ExtensionSet.gitHubFlavored));
}
```

The expected HTML output is:

`<p><strong><del>first</del></strong> <del><strong>second</strong></del></p>`

But the actual output is:

`<p><strong><del>first</del></strong> ~~<strong>second</strong>~~</p>`

This behavior happens with both bold and italic emphasis.

PR #407, opened 2022-04-05
Create DelimiterSyntax to replace TagSyntax

chenzhiguang:
Introduces a new syntax `DelimiterSyntax` which has the same parsing strategy as `TagSyntax`, but more flexible.  

For example:
```dart
const text = '#####tag#####';

final result = markdownToHtml(text, inlineSyntaxes: [
  DelimiterSyntax('#+', requiresDelimiterRun: true, tags: [
    DelimiterTag('flash', 5),
  ]),
]
```
output:

```html
<p><flash>tag</flash></p>
```

This PR also add anther syntax `EmphasisSyntax`, which replaces `TagSyntax(r'\*+', requiresDelimiterRun: true)` and `TagSyntax(r'_+', requiresDelimiterRun: true)`

---

+ Fixes https://github.com/dart-lang/markdown/issues/357


kevmoo:
This is going to need a changelog entry – right?

chenzhiguang:
> This is going to need a changelog entry – right?

Yes

chenzhiguang:
Sorry to keep committing new stuffs to this PR.
Now I am done. I will create another PR if I find any new issues or want to commit anything new.

chenzhiguang:
Sorry to keep committing new stuffs to this PR.
Now I am done. I will create another PR if I find any new issues or want to commit anything new.

lib/src/inline_parser.dart
```diff
@@ -972,7 +1016,7 @@ class LinkSyntax extends TagSyntax {
   @override
   Node? close(
       InlineParser parser, covariant SimpleDelimiter opener, Delimiter? closer,
-      {required List<Node> Function() getChildren}) {
+      {String? tag, required List<Node> Function() getChildren}) {
```
srawlins, line 1019: Style nit: please add a trailing comma here, so that each parameter appears on its own line.

lib/src/inline_parser.dart
```diff
@@ -938,26 +958,50 @@ class TagSyntax extends InlineSyntax {
...
+            allowIntraWord: true,
+            tags: [DelimiterTag('del', 2)]);
+}
 
-  @override
-  Node close(InlineParser parser, Delimiter opener, Delimiter closer,
-      {required List<Node> Function() getChildren}) {
-    return Element('del', getChildren());
-  }
+/// Parses `==mark==` to `<mark>mark</mark>`.
```
srawlins, line 988: Can you add some documentation about where this is coming from? We need specifications which we can follow in order to support special syntaxes.

lib/src/inline_parser.dart
```diff
@@ -885,16 +899,19 @@ class TagSyntax extends InlineSyntax {
   /// it on strikethrough.
   final bool allowIntraWord;
 
-  /// Create a new [TagSyntax] which matches text on [pattern].
+  final List<DelimiterTag>? tags;
+
+  /// Create a new [DelimiterSyntax] which matches text on [pattern].
```
srawlins, line 904: Nit: "Creates" instead of "Create"

https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

lib/src/inline_parser.dart
```diff
@@ -871,9 +875,19 @@ class DelimiterRun implements Delimiter {
       'canClose: $canClose>';
 }
 
+class DelimiterTag {
+  DelimiterTag(this.tag, this.indicators);
+
+  // Tag name of AST Element
```
srawlins, line 881: Grammar nit: "Tag name of the HTML element." (ending with a period)

lib/src/inline_parser.dart
```diff
@@ -871,9 +875,19 @@ class DelimiterRun implements Delimiter {
...
+  final String tag;
+
+  /// The length of [indicators]
+  final int indicators;
```
srawlins, line 885: The name of this field is probably not the best (as seen by the doc comment above it).

I would expect a field named 'indicators' to the String perhaps, or a List. This should be renamed to something like... 'indicatorLength'. If that is too long, I'd be open to other ideas...

lib/src/inline_parser.dart
```diff
@@ -938,26 +958,50 @@ class TagSyntax extends InlineSyntax {
   /// [getChildren], in which [parser] parses any nested text into child nodes.
   /// The returned [Node] incorpororates these child nodes.
   Node? close(InlineParser parser, Delimiter opener, Delimiter closer,
-      {required List<Node> Function() getChildren}) {
-    var strong = opener.length >= 2 && closer.length >= 2;
-    return Element(strong ? 'strong' : 'em', getChildren());
+      {required String tag, required List<Node> Function() getChildren}) {
```
srawlins, line 961: Style nit: please add a trailing comma here, so that each parameter appears on its own line.

srawlins (changes requested):
Note that this is a breaking change, even leaving TagSyntax intact: Some fields now return DelimiterSyntax instead of TagSyntax, and StrikethroughSyntax no longer extends TagSyntax.

If you like, you can add a CHANGELOG entry, and bump the major version number in pubspec.yaml and lib/src/version.dart.

lib/src/inline_parser.dart
chenzhiguang: Fixed

lib/src/inline_parser.dart
chenzhiguang: I have changed it to indicatorLength. Yes, indicatorLength is too long, this was the reason why I named it indicators.

lib/src/inline_parser.dart
chenzhiguang: Fixed. I thought we prefer to use this no trailing comma style.

lib/src/inline_parser.dart
chenzhiguang: Thanks! Fixed

lib/src/inline_parser.dart
chenzhiguang: Fixed

lib/src/inline_parser.dart
chenzhiguang: I did not find a common used specification about it.

My idea is all the syntaxes extend `DelimiterSyntax`  with `requiresDelimiterRun = true` follow the commonmark specification of **Emphasis and strong emphasis**: https://spec.commonmark.org/0.30/#emphasis-and-strong-emphasis

chenzhiguang:
> Note that this is a breaking change, even leaving TagSyntax intact: Some fields now return DelimiterSyntax instead of TagSyntax, and StrikethroughSyntax no longer extends TagSyntax.
> 
> If you like, you can add a CHANGELOG entry, and bump the major version number in pubspec.yaml and lib/src/version.dart.

Done!

I have anther breaking change which could be a good one. So I hope this major version release can wait for it if it is possible.

lib/src/inline_parser.dart
srawlins: I actually don't have a strong preference. Except when we have some long parameters (like with the required keyword), and now they're creeping onto a third line. In the analyzer codebase we've been moving towards using trailing commas more. It's not a hard rule yet.

chenzhiguang:
Will we merge this PR? I'm waiting for the merge to continue with other tasks to avoid a lot of conflicts. Thanks!

chenzhiguang:
Will we merge this PR? I'm waiting for the merge to continue with other tasks to avoid a lot of conflicts. Thanks!

lib/src/inline_parser.dart
srawlins: I don't think we can support an unspecified, or loosely specified syntax. Even if it is not a complicated one, it adds maintenance burden, and can create breaking changes when a well-specified, popular syntax _does_ emerge, and is slightly different from what we've implemented. As is noted in the issue, some implementations seem to support different delimiters for highlighting, so I think this is too under-specified to accept.

I'll comment more on the issue.

srawlins (changes requested):
I think that DelimiterSyntax is a fine improvement over TagSyntax, and I think we can definitely merge this PR with that feature.

But I don't think the HighlightSyntax should be a part of all this. The DelimiterSyntax definitely makes HighlightSyntax easier, so that others can implement this syntax, and perhaps other extensions for other Markdown flavors.

chenzhiguang:
Thanks! HighlightSyntax has been removed.

Yes, I agree, we should only focus on the standard  syntaxes / flavors.

srawlins (approved):
Very nice feature, thanks!

PR #412, opened 2022-04-11
Proposal: Stricter lint rules and split big files into smaller ones

chenzhiguang:
1. Enable these lint rules:
    ```
    - prefer_final_locals
    - use_if_null_to_convert_nulls_to_bools
    - prefer_interpolation_to_compose_strings
    - require_trailing_commas
    - use_raw_strings
    - unnecessary_raw_strings
    - sort_constructors_first
    ```

2.  Split inline_parser.dart and block_parser.dart into small files, each syntax a file.


srawlins:
Thanks for the proposal. I think these two ideas area probably each a good direction.

That being said, I don't think they should land as part of one change. They don't relate to each other.

In fact, I cannot really review a change to split out a few thousand lines, with the knowledge that there are definitely changes in those lines.

I would be happy to review this as separate PRs.

chenzhiguang:
Cool! Just make sure this is the right direction.

I will create separate PRs for each change when my other two ongoing PRs are closed.

chenzhiguang:
Close it now. Will create  separate PRs later.

Issue #369, opened 2021-05-17
Provide offset/lengths for nodes


DanTup:
There are some places in the analysis server that could benefit from parsing Markdown. For example https://github.com/Dart-Code/Dart-Code/issues/3330 relates to code blocks no longer being highlighted in VS Code when Semantic Tokens are enabled (Semantic Tokens replace the old regex-based colouring, but the server doesn't produce highlights regions for the code blocks nested inside markdown).

Parsing the comments using this library produces `code` elements for the things I'd like to highlight, but it appears the `offset`/`length` are not provided on the nodes so it's not possible to produce a highlight region for that code.

If this seems like a reasonable extension, I'm happy to have a go at implementing it if someone is happy to provide pointers and/or reviews. Thanks!

DanTup:
Neat! I did wonder if the current AST might cause issues for non-HTML uses, so this sounds ideal. Thanks!

frmatthew:
Any update on this issue? Thanks.

srawlins:
My teammate and I with whom I planned to work on this had a lot of non-overlapping vacation this summer, so I didn't get to land anything in June as I hoped. My new target is end of September. 🤞

DanTup:
@srawlins out of interest, is this still on your radar?

srawlins:
Oof, it is, but it keeps getting bumped by other slightly higher priority projects. :/ I should not announce a new target date.

DanTup:
lol, I know the feeling. It's not urgent to me, it's just useful to know if it may be on the way in the near-term or not :-) Thanks!

Number-3434:
2024 - Any updates yet?

srawlins:
2024 - No updates.

I would appreciate it as well. Sorry, it has not made it to the top of our priorities.

Number-3434:
Ok :(

Maybe [Dart-Code: Comment References](https://github.com/Dart-Code/Dart-Code/issues/3648)
could make it a higher priority?

Nearly all documentation comments in Dart uses some sort of comment reference, and hovering over the reference should link to the reference.

This feature is not possible without #369, so I therefore believe this should get higher priority and more attention.

Thanks!

PR #413, opened 2022-04-12
Proposal: AST enhancement

chenzhiguang:
*No need to merge this PR*

A new AST node type `Helper`(there might be a better name) is introduced, which is used to save the information other than the textContent of Text node and Element node, including markers, whitespaces, blank lines, and line feeds etc. It means the parser will not filter out nor alter any information from the input,  every character will be parsed and kept on AST.

For example:

```
  #   Foo  
```

will be parsed to a AST like this:

Screenshot 2022-04-12 at 16 19 57

With this enhancement, this package will work for more scenarios, for example somewhere wants to keep all the string input and highlight the markers, such as markdown editors or comment extensions etc.

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -71,33 +71,26 @@ class Element implements Node {
 
 class Helper implements Node {
   /// Instantiates a marker [tag] Helper.
-  Helper.marker(String name, String marker)
+  Helper.marker(String marker)
       : tag = 'marker',
-        textContent = marker,
-        attributes = {
-          'name': name,
-        };
+        textContent = marker;
 
   /// Instantiates a whitespace [tag] Helper.
   Helper.whitespace(String whitespace)
       : tag = 'whitespace',
-        textContent = whitespace,
-        attributes = {};
+        textContent = whitespace;
 
   /// Instantiates an empty line [tag] Helper
   Helper.emptyLine(String? content)
       : tag = 'emptyline',
-        textContent = content ?? '',
-        attributes = {};
+        textContent = content ?? '';
 
   /// Instantiates an new line [tag] Helper
   Helper.newLine()
       : tag = 'newline',
-        textContent = '\n',
-        attributes = {};
+        textContent = '\n';
 
   final String tag;
-  final Map<String, String> attributes;
 
   @override
   void accept(NodeVisitor visitor) => visitor.visitHelper(this);
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -55,8 +55,55 @@ class Element implements Node {
 
   @override
   String get textContent {
-    return (children ?? []).map((child) => child.textContent).join('');
+    return (children ?? [])
+        .map((child) => child is Helper ? '' : child.textContent)
+        .join('');
   }
+
+  /// If an Element has children other than Helpers.
+  bool hasContent() {
+    if (children == null || children!.isEmpty) {
+      return false;
+    }
+    return children!.whereType<Helper>().length != children!.length;
+  }
+}
+
+class Helper implements Node {
+  /// Instantiates a marker [tag] Helper.
+  Helper.marker(String name, String marker)
+      : tag = 'marker',
+        textContent = marker,
+        attributes = {
+          'name': name,
+        };
+
+  /// Instantiates a whitespace [tag] Helper.
+  Helper.whitespace(String whitespace)
+      : tag = 'whitespace',
+        textContent = whitespace,
+        attributes = {};
+
+  /// Instantiates an empty line [tag] Helper
+  Helper.emptyLine(String? content)
+      : tag = 'emptyline',
+        textContent = content ?? '',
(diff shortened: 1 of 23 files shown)


chenzhiguang:
Should we have a look at this one now?

kevmoo:
@chenzhiguang – you should rebase first, I think!

chenzhiguang:
No problem!  But My plan was that if this proposal is accepted, I will close this PR and start working on developing this feature from scratch. It will be a heavy task and will probably take a month or so. So maybe it's better to create a separate branch that I can commit PRs to.

chenzhiguang:
Ok. That is bad, I  might have copied the wrong commit id and force updated. now everything is gone!!

chenzhiguang:
Sorry, I force pushed to a wrong commit id and deleted all the commits, now they are back. :)

srawlins:
Thanks for the prototype, @chenzhiguang!

I don't think we want to mix the Markdown AST nodes (like "whitespace", `  ` and "marker", `#`) in with the HTML nodes. There are cases where a line of text, which is one continuous HTML node, is broken up into multiple Markdown AST nodes. For example:

```markdown
> Block quote with **inline
> nodes**.
  > Sloppy indentation.
Lazy continuations.
```

I think it would be an unfriendly API  to represent this with one AST with children:

1. `[marker ">"]`
2. `[whitespace " "]`
3. `[Paragraph]`
  i. `[Text "Block quote with "]`
  ii. `[Strong Emphasis]`
      a. `[marker "**"]`
      b. `[inline]`

Er... well I don't know what we would do then., in a way that allows for simple rendering of the HTML, or walking the HTML tree, and simple rendering of the Markdown AST. There is also a performance consideration to storing the Helper nodes, in lists, but then skipping over them in rendering.

I also think the simple List of children nodes is not a friendly API for walking the Markdown parse tree. In order to provide a usable API, we need to have a 'AtxHeader' node with fields, `marker`, and `text`. These fields (other node types, like "Token" perhaps, would have [SourceSpan](https://pub.dev/documentation/source_span/latest/source_span/SourceSpan-class.html) fields to identify the offsets in the source text. Then we do not nodes for whitespace. I'm thinking there can be APIs for getting the indentation of a node, which uses offsets to determine whitespace. (When `atxHeader1` has offset 0 and `atxHeader1.marker` has offset 2, we can fetch the textual content at `[0, 2]`.)

The above block quote example, which gives me nightmares, I think would need to be handled with some kind of list of markers or something. Otherwise it looks like you need a syntax _graph_ yikes.

* [BlockQuote]
  * `.markers`: [`[Token ">" at [0, 1]]`, `[Token ">" at [28, 29]]`, `[Token ">" at [41, 42]]`, `[Token "" at [63, 63]]`]
  * `.children`
    1. ...

chenzhiguang:
Thanks for such detailed feedback.

About the ***saved in parsing but skipped in rendering*** issue, I think it could reduce the performance loss with this `withHelper` disabled: [Add withHelper option to document](https://github.com/dart-lang/markdown/pull/413/commits/b06bc8b66261a271dcc3fea31890aa75abd3fe37).

I have been doing some poc development on block quote and list, in short, it won't work for the nested container blocks with only the `Helper` nodes. There are some solutions in my mind I will continue digging into. Yes, the way of using offset to record tokens could be a good solution.

But if it is not allowed to have AST nodes other than HTML nodes in the tree,  I have not figured out an alternative yet.

srawlins:
> But if it is not allowed to have AST nodes other than HTML nodes in the tree, I have not figured out an alternative yet.

I think the alternative is that we first parse the text into a Markdown AST. To render to HTML, we transform that, with a visitor, to an HTML AST, and then serialize that (with a visitor) to HTML text.

chenzhiguang:
Add a new layer on top of current HTML AST  to store all the Markdown elements such as text,  tokens or what ever? Sounds good. I will dig into it.

My idea was you could see the `Helper` nodes simply as `text` nodes but with some special attributes built in, also it is easy to skip these `Helper` nodes in rendering when you do not need them.

chenzhiguang:
You might have already done some work? I guess this issue might be related? https://github.com/dart-lang/markdown/issues/369

srawlins:
I did a prototype a while ago for #369 but have not had a chance to apply it properly to the package. I don't think I would get to it for at least a few months...

chenzhiguang:
I see. I will be working on it. Maybe we should create a new branch for the related PRs?

srawlins:
I would support a branch; that sounds great

chenzhiguang:
Thanks!

chenzhiguang:
Hi sam @srawlins! Just want to let you know that I'm working on this project right now,
But before coding, I prefer to spend a few days doing more research and learning.

chenzhiguang:
Hi Sam @srawlins! Just want to let you know that I'm working on this project right now,
But before coding, I prefer to spend a few days doing more research and learning.

PR #415, opened 2022-04-15
Stricter lint rule: enable prefer_final_locals

chenzhiguang:
https://dart.dev/guides/language/effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables

> ## DO follow a consistent rule for var and final on local variables.
> Most local variables shouldn’t have type annotations and should be declared using just var or final. There are two rules in wide use for when to use one or the other:
> 
> - Use final for local variables that are not reassigned and var for those that are.
> 
> - Use var for all local variables, even ones that aren’t reassigned. Never use final for locals. (Using final for fields and top-level variables is still encouraged, of course.)
> 
> Either rule is acceptable, but pick one and apply it consistently throughout your code. That way when a reader sees var, they know whether it means that the variable is assigned later in the function.

I pick the first rule.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -31,3 +31,4 @@ linter:
     - package_api_docs
     - test_types_in_equals
     - throw_in_finally
+    - prefer_final_locals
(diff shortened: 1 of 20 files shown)


srawlins:
I support local finals as well. @kevmoo will be happy.

kevmoo:
🤘🤙👍

srawlins (approved):
Simple enough! Thanks!

PR #416, opened 2022-04-16
Enable prefer_interpolation_to_compose_strings

chenzhiguang:
see https://dart.dev/guides/language/effective-dart/usage#prefer-using-interpolation-to-compose-strings-and-values

> ...
>  it’s almost always cleaner and shorter to use interpolation

Also it is a good idea to follow a consistent rule.


srawlins (approved):
Thanks!

PR #419, opened 2022-04-17
Enable require_trailing_commas lint rule

chenzhiguang:
Add this optional trailing commas to parameter lists to get good formatting. It is also recommended in [Flutter: Using trailing commas
](https://docs.flutter.dev/development/tools/formatting#using-trailing-commas)

### DIFF
--- a/benchmark/benchmark.dart
+++ b/benchmark/benchmark.dart
@@ -56,8 +56,10 @@ String _loadFile(String name) {
 }
 
 void _printResult(String label, double time) {
-  print('$label: ${_padLeft(time.toStringAsFixed(2), 4)}ms '
-      "${'=' * ((time * 20).toInt())}");
+  print(
+    '$label: ${_padLeft(time.toStringAsFixed(2), 4)}ms '
+    "${'=' * ((time * 20).toInt())}",
+  );
 }
 
 String _padLeft(String input, int length) {
--- a/bin/markdown.dart
+++ b/bin/markdown.dart
@@ -19,16 +19,18 @@ Future<void> main(List<String> args) async {
   final parser = ArgParser()
     ..addFlag('help', negatable: false, help: 'Print help text and exit')
     ..addFlag('version', negatable: false, help: 'Print version and exit')
-    ..addOption('extension-set',
-        allowed: ['none', 'CommonMark', 'GitHubFlavored', 'GitHubWeb'],
-        defaultsTo: 'CommonMark',
-        help: 'Specify a set of extensions',
-        allowedHelp: {
-          'none': 'No extensions; similar to Markdown.pl',
-          'CommonMark': 'Parse like CommonMark Markdown (default)',
-          'GitHubFlavored': 'Parse like GitHub Flavored Markdown',
-          'GitHubWeb': 'Parse like GitHub\'s Markdown-enabled web input fields',
-        });
+    ..addOption(
+      'extension-set',
+      allowed: ['none', 'CommonMark', 'GitHubFlavored', 'GitHubWeb'],
+      defaultsTo: 'CommonMark',
+      help: 'Specify a set of extensions',
+      allowedHelp: {
+        'none': 'No extensions; similar to Markdown.pl',
+        'CommonMark': 'Parse like CommonMark Markdown (default)',
+        'GitHubFlavored': 'Parse like GitHub Flavored Markdown',
+        'GitHubWeb': 'Parse like GitHub\'s Markdown-enabled web input fields',
+      },
+    );
   final results = parser.parse(args);
 
   if (results['help'] as bool) {
@@ -66,7 +68,8 @@ Future<void> main(List<String> args) async {
 }
 
 void printUsage(ArgParser parser) {
-  print('''Usage: markdown.dart [options] [file]
+  print(
+    '''Usage: markdown.dart [options] [file]
 
 Parse [file] as Markdown and print resulting HTML. If [file] is omitted,
 use stdin as input.
@@ -75,5 +78,6 @@ By default, CommonMark Markdown will be parsed. This can be changed with
 the --extensionSet flag.
 
 ${parser.usage}
-''');
+''',
+  );
 }
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -53,7 +53,8 @@ final _olPattern =
 
 /// A line of hyphens separated by at least one pipe.
 final _tablePattern = RegExp(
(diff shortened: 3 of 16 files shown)


kevmoo:
I like this lint, but since @srawlins is the primary maintainer here, I'll let him comment!

benchmark/benchmark.dart
```diff
@@ -56,8 +56,10 @@ String _loadFile(String name) {
 }
 
 void _printResult(String label, double time) {
-  print('$label: ${_padLeft(time.toStringAsFixed(2), 4)}ms '
-      "${'=' * ((time * 20).toInt())}");
+  print(
+    '$label: ${_padLeft(time.toStringAsFixed(2), 4)}ms '
+    "${'=' * ((time * 20).toInt())}",
```
srawlins, line 61: I am not a fan of this lint rule. For cases like this.

I am a fan of adding trailing commas to argument lists when there is more than one argument, and they span more than 1 line without the comma.

This change here is harder for me to read though.

bin/markdown.dart
```diff
@@ -19,16 +19,18 @@ Future<void> main(List<String> args) async {
...
-        allowed: ['none', 'CommonMark', 'GitHubFlavored', 'GitHubWeb'],
-        defaultsTo: 'CommonMark',
-        help: 'Specify a set of extensions',
-        allowedHelp: {
-          'none': 'No extensions; similar to Markdown.pl',
-          'CommonMark': 'Parse like CommonMark Markdown (default)',
-          'GitHubFlavored': 'Parse like GitHub Flavored Markdown',
-          'GitHubWeb': 'Parse like GitHub\'s Markdown-enabled web input fields',
-        });
+    ..addOption(
```
srawlins, line 22: I like this change though.

bin/markdown.dart
```diff
@@ -75,5 +78,6 @@ By default, CommonMark Markdown will be parsed. This can be changed with
 the --extensionSet flag.
 
 ${parser.usage}
-''');
+''',
+  );
```
srawlins, line 82: Again, I don't think this is helpful.

srawlins (commented):
I like many of the changes that this induces, but I strongly dislike some of them. What do you say to keeping many of these changes (I can mark the ones to revert), and we do not change analysis_options.yaml?

chenzhiguang:
> I like many of the changes that this induces, but I strongly dislike some of them. What do you say to keeping many of these changes (I can mark the ones to revert), and we do not change analysis_options.yaml?

I do not like some of them too. 

Yep. let me remove the rule from analysis_options.yaml, and reformat it.

lib/src/block_parser.dart
```diff
@@ -53,7 +53,8 @@ final _olPattern =
 
 /// A line of hyphens separated by at least one pipe.
 final _tablePattern = RegExp(
-    r'^[ ]{0,3}\|?([ \t]*:?\-+:?[ \t]*\|)+([ \t]|[ \t]*:?\-+:?[ \t]*)?$');
+  r'^[ ]{0,3}\|?([ \t]*:?\-+:?[ \t]*\|)+([ \t]|[ \t]*:?\-+:?[ \t]*)?$',
```
srawlins, line 56: I think this should be reverted.

lib/src/inline_parser.dart
```diff
@@ -1103,10 +1152,12 @@ class LinkSyntax extends DelimiterSyntax {
       // Normally, label text does not get parsed as inline Markdown. However,
       // for the benefit of the link resolver, we need to at least escape
       // brackets, so that, e.g. a link resolver can receive `[\[\]]` as `[]`.
-      final resolved = linkResolver(label
-          .replaceAll(r'\\', r'\')
-          .replaceAll(r'\[', '[')
-          .replaceAll(r'\]', ']'));
+      final resolved = linkResolver(
```
srawlins, line 1155: I think this should be reverted. It just feels very strange to me in cases like this where there is only one parameter...

tool/stats_lib.dart
```diff
@@ -15,18 +15,20 @@ import '../test/util.dart';
 
 // Locate the "tool" directory. Use mirrors so that this works with the test
 // package, which loads this suite into an isolate.
-String get toolDir =>
-    p.dirname((reflect(loadCommonMarkSections) as ClosureMirror)
-        .function
-        .location!
-        .sourceUri
-        .path);
+String get toolDir => p.dirname(
```
srawlins, line 18: I think this should be reverted.

But this change I think indicates a smell; this arrow-function should probably be a function body with a local variable, like a variable for the `reflect` result...

tool/update_emojis.dart
```diff
@@ -14,18 +14,25 @@ Future<void> main() async {
...
-          await response.cast<List<int>>().transform(utf8.decoder).join(''))
-      .map((String alias, dynamic info) =>
-          MapEntry(alias, info.cast<String, dynamic>()))
+    await response.cast<List<int>>().transform(utf8.decoder).join(''),
+  )
+      .map(
+        (String alias, dynamic info) => MapEntry(
+          alias,
+          info.cast<String, dynamic>(),
+        ),
```
srawlins, line 23: I think this should be reverted.

tool/update_emojis.dart
```diff
@@ -14,18 +14,25 @@ Future<void> main() async {
...
+  final emojisContent = StringBuffer(
+    '''
 // GENERATED FILE. DO NOT EDIT.
 //
 // This file was generated from emojilib's emoji data file:
 // $_emojisJsonRawUrl
 // at ${DateTime.now()} by the script, tool/update_emojis.dart.
 
-''');
+''',
```
srawlins, line 34: I think this should be reverted.

tool/update_emojis.dart
```diff
@@ -39,7 +46,9 @@ Future<void> main() async {
   });
   emojisContent.writeln('};');
   File(_emojisFilePath).writeAsStringSync(emojisContent.toString());
-  print('Wrote data to $_emojisFilePath for $emojiCount emojis, '
-      'ignoring ${ignored.length}: ${ignored.join(', ')}.');
+  print(
+    'Wrote data to $_emojisFilePath for $emojiCount emojis, '
+    'ignoring ${ignored.length}: ${ignored.join(', ')}.',
```
srawlins, line 51: I think this should be reverted.

srawlins (changes requested):
The vast majority of the changes improve readability I think.

Just a few where I think the change is much worse.

chenzhiguang:
How about close this PR?  I have just created a new [PR](https://github.com/dart-lang/markdown/pull/419) which has much fewer changes. please have a look.

srawlins:
Closing in favor of #420

PR #421, opened 2022-04-19
Enable sort_constructors_first lint rule

chenzhiguang:
It is a [Flutter style recommendation](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#constructors-come-first-in-a-class). I think it also works in Dart.


srawlins (commented):
In code owned by the Dart team, it is actually more idiomatic to put fields first. I think we should not land this one until that changes.

chenzhiguang:
Thanks. So let's skip this one for now.

PR #424, opened 2022-04-28
Remove Element.empty() from AST

chenzhiguang:
Remove `Element.empty()` which instantiates a **self-closing** tag Element and add a property `selfClosing` to represent if it is a  **self-closing** tag. 

Because if this [AST enhancement proposal]( https://github.com/dart-lang/markdown/pull/413
) is accepted, the `children` of `Element` will never be empty, it will at least have `Helper` child.

Also the current naming is ambiguous, for example:
```dart
/// Whether this element is self-closing.
bool get isEmpty => children == null;
 ```
as there are also some scenarios that need to check if `children` is empty

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -16,38 +16,38 @@ abstract class Node {
 /// A named tag that can contain other nodes.
 class Element implements Node {
   final String tag;
-  final List<Node>? children;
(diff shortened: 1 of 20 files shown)


chenzhiguang:
No need to review this PR, I will close it later.

PR #426, opened 2022-05-04
Remove dependency on package:charcode

natebosch:
Inline all used character constants into `util.dart`.

Update pubspec to modern conventions - drop the author field and change
the homepage to repository.


natebosch:
Was apparently on an old checkout and this has already been done.

PR #431, opened 2022-05-19
Fix visitText to use Text.textContent getter instead of a Text.text member variable

timmaffett:
This PR is a simple change to have `visitText()` use the `.textContent` getter of the AST `Text` class instead of accessing the `.text` member variable directly.  This allows proper behavior of `visitText()` when processing nodes which are subclasses `Text`.
(The `textContent` getter of `Text` is already a overridden getter from `Element`, accessing the text content via that getter seems to be the correct way).
 
(more background details:)
In making a block syntax extension that is capable of asynchronous modifications to code block nodes I needed to create asynchronous text nodes who's values finalized after the time of creation.  (and `.text` is a final variable... overriding it did not seem like a clean approach)..


srawlins:
Cool PR, thanks. Is it possible to add a test case for such a custom extension, or is it a ton of code?

timmaffett:
I was planning on submitting a PR with the async transformation as I was thinking that it may be generally useful.  I will wrap it up and submit it today.  If nothing else I would love feedback on the approach - there is not much code at all really.

(background:)
I created it to support using the Kroki web service to allow for 'compile time' support of mermaid (and other) diagram types by dartdoc, etc.  Being able to statically convert diagram code blocks to svgs at the time of markdown conversion seems to be the better approach than creating the diagrams at runtime on the client side (for security considerations and otherwise).   Kroki is a open source (https://github.com/yuzutech/kroki) (https://kroki.io) web service for doing static conversions of various diagram formats and seems well suited for the dartdoc/pub.dev support of diagrams, as the conversion would be done at the time of doc creation leaving only static inline svgs.  
Your original on [comments](https://github.com/dart-lang/markdown/issues/228) about mermaid diagrams started me on this journey and I think this static conversion may be the approach you were imagining.

timmaffett:
@srawlins Is there any chance this PR could get included in 6.0.1 ?

srawlins:
This seems like a valid fix

srawlins (approved):

PR #441, opened 2022-08-08
Footnote support

lindeer:
implementation referred to [markdown-it-footnote](https://github.com/markdown-it/markdown-it-footnote/blob/master/index.js) and [pandoc](https://pandoc.org/MANUAL.html#footnotes)

### DIFF
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -6,9 +6,9 @@ import 'block_syntax.dart' show BlockSyntax;
 
 /// Footnote definition could contain multiple line-children and children could
 /// be separated by one empty line.
-/// Its first child-line would be the remaining part of first line after taking
-/// definition leading, combining with other line-children parsed by
-/// `parseChildLines`, is feed into `BlockParser`.
+/// Its first child-line would be the remaining part of the first line after
+/// taking definition leading, combining with other child lines parsed by
+/// [parseChildLines], is fed into [BlockParser].
 class FootnoteDefSyntax extends BlockSyntax {
   const FootnoteDefSyntax();
 
@@ -29,8 +29,6 @@ class FootnoteDefSyntax extends BlockSyntax {
       Line(current.substring(match[0]!.length)),
       ...parseChildLines(parser),
     ];
-    // Linebreak in secondary paragraph did not match LineBreakSyntax
-    // so there is no `<br />`, unlike GitHub.
     final children = BlockParser(lines, parser.document).parseLines();
     return Element('li', children)
       ..attributes['id'] = 'fn-$id'
@@ -42,11 +40,8 @@ class FootnoteDefSyntax extends BlockSyntax {
     final children = <String>[];
     // As one empty line should not split footnote definition, use this flag.
     var shouldBeBlock = false;
-    Iterable<BlockSyntax>? syntaxList;
-    Iterable<BlockSyntax> validSyntaxList() {
-      return syntaxList ??= parser.blockSyntaxes
-          .where((s) => !_excludingPattern.contains(s.pattern));
-    }
+    late final syntaxList = parser.blockSyntaxes
+        .where((s) => !_excludingPattern.contains(s.pattern));
 
     while (!parser.isDone) {
       final line = parser.current.content;
@@ -59,7 +54,7 @@ class FootnoteDefSyntax extends BlockSyntax {
         children.add(line.substring(4));
         parser.advance();
         shouldBeBlock = false;
-      } else if (shouldBeBlock || _isBlock(validSyntaxList(), line)) {
+      } else if (shouldBeBlock || _isBlock(syntaxList, line)) {
         break;
       } else {
         children.add(line);
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -43,7 +43,7 @@ class FootnoteDefSyntax extends BlockSyntax {
     late final syntaxList = parser.blockSyntaxes
         .where((s) => !_excludingPattern.contains(s.pattern));
 
-    // Every line is footnote's children util two blank lines or a block
+    // Every line is footnote's children util two blank lines or a block.
     while (!parser.isDone) {
       final line = parser.current.content;
       if (line.trim().isEmpty) {
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -65,13 +65,13 @@ class FootnoteDefSyntax extends BlockSyntax {
     return children.map(Line.new).toList(growable: false);
   }
 
-  // Patterns that would be used to decide if one line is a block
+  /// Patterns that would be used to decide if one line is a block.
   static final _excludingPattern = {
     emptyPattern,
     dummyPattern,
   };
 
-  // Whether this line is one kind of block, if true footnotes block should end.
+  /// Whether this line is one kind of block, if true footnotes block should end.
   static bool _isBlock(Iterable<BlockSyntax> syntaxList, String line) {
     return syntaxList.any((s) => s.pattern.hasMatch(line));
   }
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -43,6 +43,7 @@ class FootnoteDefSyntax extends BlockSyntax {
     late final syntaxList = parser.blockSyntaxes
         .where((s) => !_excludingPattern.contains(s.pattern));
 
+    // Every line is footnote's children util two blank lines or a block
     while (!parser.isDone) {
       final line = parser.current.content;
       if (line.trim().isEmpty) {
@@ -64,11 +65,13 @@ class FootnoteDefSyntax extends BlockSyntax {
     return children.map(Line.new).toList(growable: false);
   }
 
+  // Patterns that would be used to decide if one line is a block
   static final _excludingPattern = {
     emptyPattern,
     dummyPattern,
   };
 
+  // Whether this line is one kind of block, if true footnotes block should end.
   static bool _isBlock(Iterable<BlockSyntax> syntaxList, String line) {
     return syntaxList.any((s) => s.pattern.hasMatch(line));
   }
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -1,11 +1,8 @@
 import '../ast.dart' show Element, Node;
 import '../block_parser.dart' show BlockParser;
 import '../line.dart';
-import '../patterns.dart' show footnotePattern;
+import '../patterns.dart' show dummyPattern, emptyPattern, footnotePattern;
 import 'block_syntax.dart' show BlockSyntax;
-import 'empty_block_syntax.dart';
-import 'paragraph_syntax.dart';
-import 'setext_header_syntax.dart';
 
 /// Footnote definition could contain multiple line-children and children could
 /// be separated by one empty line.
@@ -37,7 +34,7 @@ class FootnoteDefSyntax extends BlockSyntax {
     final children = BlockParser(lines, parser.document).parseLines();
     return Element('li', children)
       ..attributes['id'] = 'fn-$id'
-      ..attributes['_label_'] = label;
+      ..footnoteLabel = label;
   }
 
   @override
@@ -48,7 +45,7 @@ class FootnoteDefSyntax extends BlockSyntax {
     Iterable<BlockSyntax>? syntaxList;
     Iterable<BlockSyntax> validSyntaxList() {
       return syntaxList ??= parser.blockSyntaxes
-          .where((s) => !_invalidSecondaryBlock.contains(s.runtimeType));
+          .where((s) => !_excludingPattern.contains(s.pattern));
     }
 
     while (!parser.isDone) {
@@ -72,10 +69,9 @@ class FootnoteDefSyntax extends BlockSyntax {
     return children.map(Line.new).toList(growable: false);
   }
 
-  static const _invalidSecondaryBlock = <Type>{
-    EmptyBlockSyntax,
-    ParagraphSyntax,
-    SetextHeaderSyntax,
+  static final _excludingPattern = {
+    emptyPattern,
+    dummyPattern,
   };
 
   static bool _isBlock(Iterable<BlockSyntax> syntaxList, String line) {
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -1,0 +1,73 @@
+import '../ast.dart' show Element, Node;
+import '../block_parser.dart' show BlockParser;
+import '../patterns.dart' show footnotePattern;
+import 'block_syntax.dart' show BlockSyntax;
+import 'empty_block_syntax.dart';
+import 'paragraph_syntax.dart';
+import 'setext_header_syntax.dart';
+
+class FootnoteDefSyntax extends BlockSyntax {
+  const FootnoteDefSyntax();
+
+  @override
+  RegExp get pattern => footnotePattern;
+
+  @override
+  Node? parse(BlockParser parser) {
+    final current = parser.current;
+    final match = pattern.firstMatch(current)!;
+    final label = match[2]!;
+    final refs = parser.document.footnoteReferences;
+    refs[label] = 0;
+
+    final id = Uri.encodeComponent(label);
+    parser.advance();
+    final lines = [
+      current.substring(match[0]!.length),
+      ...parseChildLines(parser),
+    ];
+    // Linebreak in secondary paragraph did not match LineBreakSyntax
+    // so there is no `<br />`, unlike github.
+    final children = BlockParser(lines, parser.document).parseLines();
+    return Element('li', children)
+      ..attributes['id'] = 'fn-$id'
+      ..attributes['_label_'] = label;
+  }
+
+  @override
+  List<String> parseChildLines(BlockParser parser) {
+    final children = <String>[];
+    var nextIsBlock = false;
+    while (!parser.isDone) {
+      final line = parser.current;
+      if (line.trim().isEmpty) {
+        children.add(line);
+        parser.advance();
+        nextIsBlock = true;
+        continue;
+      } else if (line.startsWith('    ')) {
+        children.add(line.substring(4));
+        parser.advance();
+        nextIsBlock = false;
+      } else if (nextIsBlock || _isBlock(parser, line)) {
+        break;
+      } else {
+        children.add(line);
+        parser.advance();
+      }
+    }
+    return children;
+  }
+
+  static const _invalidSecondaryBlock = <Type>[
+    EmptyBlockSyntax,
+    ParagraphSyntax,
+    SetextHeaderSyntax,
+  ];
+
+  static bool _isBlock(BlockParser parser, String line) {
+    final syntaxList = parser.blockSyntaxes
+        .takeWhile((s) => !_invalidSecondaryBlock.contains(s.runtimeType));
+    return syntaxList.any((s) => s.pattern.hasMatch(line));
+  }
+}
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -4,6 +4,9 @@ import '../line.dart';
 import '../patterns.dart' show dummyPattern, emptyPattern, footnotePattern;
 import 'block_syntax.dart' show BlockSyntax;
 
+/// The spec of GFM about footnotes is [missing](https://github.com/github/cmark-gfm/issues/283#issuecomment-1378868725).
+/// For online source code of cmark-gfm, see [master@c32ef78](https://github.com/github/cmark-gfm/blob/c32ef78/src/blocks.c#L1212).
+/// A Rust implementation is also [available](https://github.com/wooorm/markdown-rs/blob/2498e31eecead798efc649502bbf5f86feaa94be/src/construct/gfm_footnote_definition.rs).
 /// Footnote definition could contain multiple line-children and children could
 /// be separated by one empty line.
 /// Its first child-line would be the remaining part of the first line after
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -7,6 +7,11 @@ import 'empty_block_syntax.dart';
 import 'paragraph_syntax.dart';
 import 'setext_header_syntax.dart';
 
+/// Footnote definition could contain multiple line-children and children could
+/// be separated by one empty line.
+/// Its first child-line would be the remaining part of first line after taking
+/// definition leading, combining with other line-children parsed by
+/// `parseChildLines`, is feed into `BlockParser`.
 class FootnoteDefSyntax extends BlockSyntax {
   const FootnoteDefSyntax();
 
@@ -28,7 +33,7 @@ class FootnoteDefSyntax extends BlockSyntax {
       ...parseChildLines(parser),
     ];
     // Linebreak in secondary paragraph did not match LineBreakSyntax
-    // so there is no `<br />`, unlike github.
+    // so there is no `<br />`, unlike GitHub.
     final children = BlockParser(lines, parser.document).parseLines();
     return Element('li', children)
       ..attributes['id'] = 'fn-$id'
@@ -38,19 +43,26 @@ class FootnoteDefSyntax extends BlockSyntax {
   @override
   List<Line> parseChildLines(BlockParser parser) {
     final children = <String>[];
-    var nextIsBlock = false;
+    // As one empty line should not split footnote definition, use this flag.
+    var shouldBeBlock = false;
+    Iterable<BlockSyntax>? syntaxList;
+    Iterable<BlockSyntax> validSyntaxList() {
+      return syntaxList ??= parser.blockSyntaxes
+          .where((s) => !_invalidSecondaryBlock.contains(s.runtimeType));
+    }
+
     while (!parser.isDone) {
       final line = parser.current.content;
       if (line.trim().isEmpty) {
         children.add(line);
         parser.advance();
-        nextIsBlock = true;
+        shouldBeBlock = true;
         continue;
       } else if (line.startsWith('    ')) {
         children.add(line.substring(4));
         parser.advance();
-        nextIsBlock = false;
-      } else if (nextIsBlock || _isBlock(parser, line)) {
+        shouldBeBlock = false;
+      } else if (shouldBeBlock || _isBlock(validSyntaxList(), line)) {
         break;
       } else {
         children.add(line);
@@ -60,15 +72,13 @@ class FootnoteDefSyntax extends BlockSyntax {
     return children.map(Line.new).toList(growable: false);
   }
 
-  static const _invalidSecondaryBlock = <Type>[
+  static const _invalidSecondaryBlock = <Type>{
     EmptyBlockSyntax,
     ParagraphSyntax,
     SetextHeaderSyntax,
-  ];
+  };
 
-  static bool _isBlock(BlockParser parser, String line) {
-    final syntaxList = parser.blockSyntaxes
-        .takeWhile((s) => !_invalidSecondaryBlock.contains(s.runtimeType));
+  static bool _isBlock(Iterable<BlockSyntax> syntaxList, String line) {
     return syntaxList.any((s) => s.pattern.hasMatch(line));
   }
 }
--- a/lib/src/block_syntaxes/footnote_def_syntax.dart
+++ b/lib/src/block_syntaxes/footnote_def_syntax.dart
@@ -1,5 +1,6 @@
 import '../ast.dart' show Element, Node;
 import '../block_parser.dart' show BlockParser;
+import '../line.dart';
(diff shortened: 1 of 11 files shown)


lindeer:
hi @srawlins , could you please have a review for footnote support?

chenzhiguang:
I created a footnote reference specification, it has some slightly different from some of the implementations. But at least there are clear rules: 🙂

https://github.com/chenzhiguang/dart_markdown/wiki/footnote-reference-specification

chenzhiguang:
Test cases:
https://github.com/chenzhiguang/dart_markdown/blob/master/test/extensions/footnote_reference.unit

lindeer:
thanks @chenzhiguang , I think footnote is an import markdown feature, especially in article writing though not in spec, not understand why official library do not support it, nearly every other language md library has it!  But I find your implementation has huge difference from markdown-it and pandoc, and your based dart-lang md library has also great changes. I have a little concern for future maintenance.

srawlins:
I think the primary spec we would follow (barring commonmark, since that does not support footnotes) would be GitHub (via https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/). I do really like pandoc, and think it is a popular project. I have not heard of markdown-it. So if we land an implementation it will have to be compatible with GitHub with a good array of tests.

The good array of tests is needed since GitHub does not have a spec for their footnote support (😢 ). That is still a bummer because we just have to... manually test with Babelmark or something? 😢

lindeer:
Fine @srawlins , I would refine the code according to github's [example](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes). The main difference is paragraph making in footnote-definition part, and I will add more test cases.

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,32 @@
+import '../ast.dart';
+import '../charcode.dart';
+import '../inline_parser.dart';
+import 'inline_syntax.dart';
+
+class FootnoteRefSyntax extends InlineSyntax {
+  FootnoteRefSyntax() : super(r'!?\[\^(\S+?)\]');
```
srawlins, line 7: Hmm, what is the leading `!?` for?

test/extensions/footnote_block.unit
```diff
@@ -0,0 +1,80 @@
+>>> footnote with paragraph
```
srawlins, line 1: Also it looks like leading whitespace is allowed in the ref?

```markdown
Footnote 3 link[^ a].

[^a]: Text
```

works in GitHub.

srawlins, line 1: Can we add a test where the footnote def is first?

```markdown
[^a]: Text

Footnote 3 link[^a].
```

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  @override
+  bool onMatch(InlineParser parser, Match match) {
+    final label = match[1]!;
+    final refs = parser.document.footnoteReferences;
+    if (!refs.containsKey(label)) {
```
srawlins, line 13: Since `footnoteReferences` is typed with non-nullable types, we can simplify some of this code by grabbing the value first with `final something = parser.document.footnoteReferences[label];`. Then we can guard this if-statement with `if (something != null)` and line 19 can be `parser.document.footnoteReferences[label] = something + 1;`

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,32 @@
...
+      final start = parser.start;
+      final text = parser.source.substring(start, start + match[0]!.length);
+      parser.addNode(Text(text));
+      return true;
+    }
+    refs[label] = (refs[label] ?? 0) + 1;
+    final pos = refs.keys.toList(growable: false).indexOf(label);
+
+    if (parser.charAt(parser.start) == $exclamation) {
```
srawlins, line 22: Why is an exclamation handled separately here?

lib/src/document.dart
```diff
@@ -65,6 +66,8 @@ class Document {
   List<Node> parseLines(List<String> lines) {
     final nodes = BlockParser(lines, this).parseLines();
     _parseInlineContent(nodes);
+    // after parsing inline as we need ref count
```
srawlins, line 69: Make sure to start comments with a capiital, and end with a period. https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

lib/src/document.dart
```diff
@@ -84,6 +87,41 @@ class Document {
...
+        }
+      } else {
+        blocks.add(node);
+      }
+    }
+    nodes
+      ..clear()
+      ..addAll(blocks);
+
+    final section = footnotes.isEmpty
```
srawlins, line 115: Can these few lines be simplified by guarding them all with `footnotes.isEmpty` rather than doing a second guard below with `section != null`.

lib/src/document.dart
```diff
@@ -84,6 +87,41 @@ class Document {
...
+        ? null
+        : _classElement('section', 'footnotes', [
+            _classElement('ol', 'footnotes-list', footnotes),
+          ]);
+    if (section != null) {
+      nodes.add(_classElement('hr', 'footnotes-divider', null));
```
srawlins, line 121: I don't think we need the `null`.

lib/src/document.dart
```diff
@@ -84,6 +87,41 @@ class Document {
       }
     }
   }
+
+  static Element _classElement(
```
srawlins, line 91: This name is a bit odd; maybe something like `_elementWithClass`.

Alternatively, this helper may not be needed with my suggestions below; `Element(tag, children)..attributes['class'] = foo` should be fine to read.

srawlins (changes requested):
Thanks for the PR!

srawlins:
@lindeer thanks for the new commit! Since it was force-pushed and overwrote the previous content, I can't see what was changed. Could you go through my comments and say something like "Done" on each if it has been addressed, and click "Resolve conversation" on those. I think I had a question or two as well which you could answer inline (I don't think I was specifically asking for code comments...) Thanks!

test/extensions/footnote_block.unit
```diff
@@ -0,0 +1,80 @@
...
+<p>"Smartypants, double quotes" and 'single quotes'</p>
+<hr class="footnotes-divider" />
+<section class="footnotes">
+<ol class="footnotes-list">
+<li id="fn1" class="footnote-item">
+<p>Footnote <strong>can have markup</strong></p>
+<p>and multiple paragraphs. <a href="#fnref1" class="footnote-backref">↩︎</a></p>
+</li>
+</ol>
+</section>
```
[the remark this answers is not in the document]
lindeer: done

lib/src/inline_syntaxes/footnote_ref_syntax.dart
lindeer: refine the code and not extends but directly change within `LinkSyntax`

test/extensions/footnote_block.unit
lindeer: now supported

lib/src/inline_syntaxes/footnote_ref_syntax.dart
lindeer: code  refined, and not need any more.

lib/src/document.dart
lindeer: done

lib/src/document.dart
lindeer: change guarding to `isEmpty`

lib/src/document.dart
lindeer: done also

lindeer:
sorry srawlins, I squashed commits too early, and github not addressed automatically. I think most important part of new commits was the introducing of 'reparse' in InlineParser. When failed to match link, we check if current content match footnote between '[' and ']', "[...][“ and "]", this result in that some leading content lost, so rollback 'pos' and 'start' to starting position and do some clearing things.  Other parts reuse link syntax logic, trying to do minimum changes.

And I added many corner cases, some different from github (I commented `unlike github`), but that's because of our own design.

lindeer:
hi @srawlins ,I found my way of 'reparse' implementation was  not elegant, we could modify `close()` of  `LinkSyntax`  without any change of `InlineParser`. So could I drop this PR and create 2 separated PRs?
1. first one for modifying existing  code without any regression.
2. after the first was  merged, the second rebased onto the first,  and was to implement footnote.

This way could make code modification more clear to review, could I?

srawlins:
@lindeer That's not a bad idea. But you don't have to close this one first. If you create a PR for modifying `LinkSyntax.close`, I can see how it would benefit this PR.

srawlins:
Hi @lindeer I see you pushed last week. Is this ready for review again? I've lost context, apologies.

lindeer:
@srawlins yes, it is. I just rebased onto new main branch, and added a new rebase fix, but github listed all commits.

lib/src/document.dart
```diff
@@ -14,6 +14,10 @@ import 'util.dart';
 /// Maintains the context needed to parse a Markdown document.
 class Document {
   final Map<String, LinkReference> linkReferences = {};
+  // keys are case sensitive.
```
srawlins, line 17: Nit: Start with a capital letter [1] and hyphenate 'case-sensitive'.

[1] https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

lib/src/document.dart
```diff
@@ -14,6 +14,10 @@ import 'util.dart';
...
+  final footnoteReferences = <String, int>{};
+  // Keep footnotes appearing order, and labels are case insensitive.
```
srawlins, line 19: Nit: Hyphenate 'case-insensitive'.

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,74 @@
+import '../ast.dart' show Element, Node;
+import '../block_parser.dart' show BlockParser;
+import '../line.dart';
+import '../patterns.dart' show footnotePattern;
+import 'block_syntax.dart' show BlockSyntax;
+import 'empty_block_syntax.dart';
+import 'paragraph_syntax.dart';
+import 'setext_header_syntax.dart';
+
+class FootnoteDefSyntax extends BlockSyntax {
```
srawlins, line 10: Would you mind adding a doc comment explaining the source of this syntax? Thanks!

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,74 @@
...
+    refs[label] = 0;
+
+    final id = Uri.encodeComponent(label);
+    parser.advance();
+    final lines = [
+      Line(current.substring(match[0]!.length)),
+      ...parseChildLines(parser),
+    ];
+    // Linebreak in secondary paragraph did not match LineBreakSyntax
+    // so there is no `<br />`, unlike github.
```
srawlins, line 31: nit: 'GitHub' should be capitalized.

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  static const _invalidSecondaryBlock = <Type>[
+    EmptyBlockSyntax,
+    ParagraphSyntax,
+    SetextHeaderSyntax,
+  ];
+
+  static bool _isBlock(BlockParser parser, String line) {
+    final syntaxList = parser.blockSyntaxes
+        .takeWhile((s) => !_invalidSecondaryBlock.contains(s.runtimeType));
```
srawlins, line 71: I don't want to use runtimeType here, as that would not account for (theoretical, to be sure) subtypes.

I think we can use `.whereType` and a late instance field to compute this once per parse. Something like:

```dart
late final secondaryBlockSyntaxes =
  [...parser.blockSyntaxes]
    .removeWhere((e) = _invalidSecondaryBlock.contains(e));
```

WDYT?

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,74 @@
...
+        break;
+      } else {
+        children.add(line);
+        parser.advance();
+      }
+    }
+    return children.map(Line.new).toList(growable: false);
+  }
+
+  static const _invalidSecondaryBlock = <Type>[
```
srawlins, line 63: Let's make this a Set since it is only used for `.contains`.

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,74 @@
...
+    final lines = [
+      Line(current.substring(match[0]!.length)),
+      ...parseChildLines(parser),
+    ];
+    // Linebreak in secondary paragraph did not match LineBreakSyntax
+    // so there is no `<br />`, unlike github.
+    final children = BlockParser(lines, parser.document).parseLines();
+    return Element('li', children)
+      ..attributes['id'] = 'fn-$id'
+      ..attributes['_label_'] = label;
```
srawlins, line 35: Instead of a fake HTML attribute which is removed later, we can add a field to Element called `footnoteLabel` or something similar. Like how Element has a `generatedId` for header IDs.

lib/src/document.dart
```diff
@@ -97,6 +103,79 @@ class Document {
...
+      } else {
+        blocks.add(node);
+      }
+    }
+    nodes
+      ..clear()
+      ..addAll(blocks);
+
+    if (footnotes.isNotEmpty) {
+      // sort footnotes by appeared order
```
srawlins, line 132: Nit: start with a capital letter and end with a period.

lib/src/document.dart
```diff
@@ -97,6 +103,79 @@ class Document {
...
+          'fn-${footnoteLabels[i]}': i,
+      };
+      footnotes.sort((l, r) {
+        final idl = l.attributes['id']?.toLowerCase() ?? '';
+        final idr = r.attributes['id']?.toLowerCase() ?? '';
+        return (ordinal[idl] ?? 0) - (ordinal[idr] ?? 0);
+      });
+      final list = Element('ol', footnotes);
+
+      // Ignore github's attribute: <data-footnotes>.
```
srawlins, line 144: Spelling, 'GitHub'.

lib/src/document.dart
```diff
@@ -97,6 +103,79 @@ class Document {
...
+        yield Text(' ');
+        final num = '${i + 1}';
+        final suffix = i > 0 ? '-$num' : '';
+        yield Element('a', [
+          Text('\u21a9'),
+          if (i > 0)
+            Element('sup', [Text(num)])..attributes['class'] = 'footnote-ref',
+        ])
+
+          // Ignore github's attributes: <data-footnote-backref aria-label="Back to content">.
```
srawlins, line 163: Spelling: 'GitHub'

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,69 @@
+import '../ast.dart' show Element, Node, Text;
+import '../charcode.dart';
+import '../inline_parser.dart' show InlineParser;
+import 'link_syntax.dart' show LinkContext;
+
+// Footnotes syntax share the same heading '[' with Reference Link, this makes
+// a lot of mass, if extends `LinkSyntax`, it will have to override
```
srawlins, line 7: This comment is fine for the PR, but it doesn't have to be checked in. It can be removed.

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,69 @@
...
+// `_tryCreateReferenceLink`, and also need other context info, so just a
+// separated class is enough.
+class FootnoteRefSyntax {
+  static String? _footnoteLabel(InlineParser parser, String key) {
```
srawlins, line 11: It doesn't look like this needs `parser`?

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,69 @@
...
+    if (key.isEmpty) {
+      return null;
+    }
+    return key;
+  }
+
+  static Iterable<Node>? tryCreateFootnoteLink(
+    LinkContext context,
+    String text,
+    bool secondary,
```
srawlins, line 25: Avoid positional bool parameters. https://dart.dev/guides/language/effective-dart/design#avoid-positional-boolean-parameters

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,69 @@
...
+  ) {
+    final parser = context.parser;
+    final key = _footnoteLabel(parser, text);
+    final refs = parser.document.footnoteReferences;
+    // `label` is what footnoteReferences stored, it is case sensitive.
+    final label =
+        refs.keys.firstWhere((k) => k.toLowerCase() == key, orElse: () => '');
+    // `count != null` means footnote was valid
```
srawlins, line 33: End comments with period, here and below.

lib/src/patterns.dart
```diff
@@ -50,6 +50,9 @@ final listPattern =
 final tablePattern = RegExp(
     r'^[ ]{0,3}\|?([ \t]*:?\-+:?[ \t]*\|)+([ \t]|[ \t]*:?\-+:?[ \t]*)?$');
 
+/// Just keep the same with github.
```
srawlins, line 53: This comment does not describe what this pattern is. We don't have to reference GitHub.

srawlins (changes requested):
Great work! I have a few comments and then I will review further. Thanks again!

lib/src/block_syntaxes/footnote_def_syntax.dart
lindeer: a new field make `Element` inflating, and it is a field hardly could be used by other subtype. What about using 'data attribute' from [HTML syntax](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes), attributes to store some extra information, just like `data-footnote-label`?

lindeer:
hi @srawlins , seems every time I pushed a new commit, the review came into a pause, hope this could be merged earlier :cry:

srawlins:
Yes, apologies. We need to figure out a different system for open source contributions.

lib/src/block_syntaxes/footnote_def_syntax.dart
srawlins: I don't think we have any memory inflation concerns for this package. It doesn't outweigh the importance of the readability of the code here. A field brings in type support, typo prevention, other forms of static analysis, possible runtime optimizations, etc.

lib/src/block_syntaxes/footnote_def_syntax.dart
srawlins: > seems have to use runtimeType, we actually expect a iterable with same base type `BlockSyntax` but different subtype, `whereType` could only give us a iterable with same type.

Sorry I don't understand this. You might be misunderstanding what `whereType` does. `[1, 2.1, 'hello'].whereType<num>()` will yield both `1` and `2.1`.

> And for a late instance, I use `??=` to assign a local variable, a class field would make `FootnoteDefSyntax` none const constructor any more.

OK a local variable is ok here.

srawlins (changes requested):

lib/src/block_syntaxes/footnote_def_syntax.dart
lindeer: I use BlockSytax's `pattern` field to exclude `EmptyBlockSyntax`, `ParagraphSyntax` and `SetextHeaderSyntax`, no `runtimeType` any more.

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,65 @@
...
+    final label =
+        refs.keys.firstWhere((k) => k.toLowerCase() == key, orElse: () => '');
+    // `count != null` means footnote was valid.
+    var count = refs[label];
+    // And then check if footnote was matched.
+    if (key == null || count == null) {
+      return null;
+    }
+    final result = <Node>[];
+    // There are 4 cases here: ![^...], [^...], ![...][^...], [...][^...]
```
srawlins, line 36: Very helpful comment

lib/src/inline_syntaxes/footnote_ref_syntax.dart
```diff
@@ -0,0 +1,65 @@
+import '../ast.dart' show Element, Node, Text;
+import '../charcode.dart';
+import 'link_syntax.dart' show LinkContext;
+
+class FootnoteRefSyntax {
```
srawlins, line 5: Could you add comments here about where this syntax is specified? Could just be live experiments from GitHub, if it is not spelled out in GitHub-flavored Syntax.

srawlins (changes requested):
This is looking so great! Thanks so much for working on this, I love it.

lib/src/inline_syntaxes/footnote_ref_syntax.dart
lindeer: It is a handling branch of link syntax, but I move the code into a separated class.

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,80 @@
...
+    refs[label] = 0;
+
+    final id = Uri.encodeComponent(label);
+    parser.advance();
+    final lines = [
+      Line(current.substring(match[0]!.length)),
+      ...parseChildLines(parser),
+    ];
+    // Linebreak in secondary paragraph did not match LineBreakSyntax
+    // so there is no `<br />`, unlike GitHub.
```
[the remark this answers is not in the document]
lindeer: sorry, it's a mistake, I ran `cmark-gfm` with `--hardbreaks`, remove the comment.

lindeer:
@srawlins could you give another review? :pray:

lib/src/document.dart
```diff
@@ -97,6 +106,81 @@ class Document {
...
+      final section = Element('section', [list])
+        ..attributes['class'] = 'footnotes';
+      blocks.add(section);
+    }
+    return blocks;
+  }
+
+  /// Generate backref nodes, append them to footnote definition's last child.
+  void _appendBackref(List<Node> children, String ref, int count) {
+    final refs = () sync* {
```
srawlins, line 157: Unfortunately `sync*` is slow. https://github.com/dart-lang/sdk/issues/32102

Why do we need a self-executing block here? (Apologies if I suggested it 😅 )

It looks like this can be:

```dart
var refs = [
  for (var i = 0; i < count; i++)
    ...[
      Text(' '),
      Element('a', [
          Text('\u21a9'),
          if (i > 0)
            Element('sup', [Text(num)])..attributes['class'] = 'footnote-ref',
        ])
            // Ignore GitHub's attributes: <data-footnote-backref aria-label="Back to content">.
              ..attributes['href'] = '#fnref-$ref$suffix'
              ..attributes['class'] = 'footnote-backref',
    ],
];
```

OK so that is _really_ not readable, haha, so I would extract out some logic into a either a real or a fake constructor on Element, like:

```dart
extension on Element {
  static Element footnoteAnchor(String ref, int index) {
    final number = '${i + 1}';
    final suffix = i > 0 ? '-$number' : '';
    return Element('a', [
          Text('\u21a9'),
          if (index > 0)
            Element('sup', [Text(number)])..attributes['class'] = 'footnote-ref',
        ])

          // Ignore GitHub's attributes: `<data-footnote-backref aria-label="Back to content">`.
          ..attributes['href'] = '#fnref-$ref$suffix'
          ..attributes['class'] = 'footnote-backref';
  }
}
```

Then the above list mess can become:


```dart
var refs = [
  for (var i = 0; i < count; i++)
    ...[Text(' '), Element.footnoteAnchor(ref, i)],
];
```

srawlins (changes requested):
Almost there, thanks!

lib/src/block_syntaxes/footnote_def_syntax.dart
srawlins: This needs to be addressed, please.

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,78 @@
...
+        break;
+      } else {
+        children.add(line);
+        parser.advance();
+      }
+    }
+    return children.map(Line.new).toList(growable: false);
+  }
+
+  // Patterns that would be used to decide if one line is a block
```
srawlins, line 68: Please use triple-slash for doc comments (see [Effective Dart](https://dart.dev/guides/language/effective-dart/documentation#do-use--doc-comments-to-document-members-and-types)) and end with a period. (see [Effective Dart](https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences)).

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,78 @@
...
+  static final _excludingPattern = {
+    emptyPattern,
+    dummyPattern,
+  };
+
+  // Whether this line is one kind of block, if true footnotes block should end.
```
srawlins, line 74: Please use `///` here too.

lib/src/inline_syntaxes/footnote_ref_syntax.dart
srawlins: Please include a more full comment here. Something like "A syntax for defining a footnote reference, as specified at..." and include a click-able link.

srawlins (changes requested):

lib/src/inline_syntaxes/footnote_ref_syntax.dart
lindeer: resolved in a8a81fc

lib/src/block_syntaxes/footnote_def_syntax.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  @override
+  List<Line> parseChildLines(BlockParser parser) {
+    final children = <String>[];
+    // As one empty line should not split footnote definition, use this flag.
+    var shouldBeBlock = false;
+    late final syntaxList = parser.blockSyntaxes
+        .where((s) => !_excludingPattern.contains(s.pattern));
+
+    // Every line is footnote's children util two blank lines or a block
```
srawlins, line 46: Nit: end with a period: https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

lib/src/block_syntaxes/footnote_def_syntax.dart
srawlins: I'm sorry, @lindeer , we cannot accept this PR unless this comment is addressed. We need to have documented where the syntax is specified. Who controls its definition / where can we go for updates, etc?

lib/src/inline_syntaxes/footnote_ref_syntax.dart
srawlins: Please address the above comment:

> Please include a more full comment here. Something like "A syntax for defining a footnote reference, as specified at..." and include a click-able link.

A click-able link would probably have a URL starting with 'https://'

lindeer: footnote reference here is a handling branch of `LinkSyntax`, not a kind of syntax. I just moved these code into a separated class, should I rename the class to avoid misunderstanding?

lib/src/block_syntaxes/footnote_def_syntax.dart
lindeer: sorry @srawlins , I have no idea why I missed these comment..

lindeer: I came to know that [footnotes are not in the spec yet](https://github.com/github/cmark-gfm/issues/283#issuecomment-1378868725).  It is said that there is a private repo for the spec somewhere in GH.
I found a GFM footnote grammar and its rust implementation [here](https://github.com/wooorm/markdown-rs/blob/2498e31eecead798efc649502bbf5f86feaa94be/src/construct/gfm_footnote_definition.rs).
or what other I could do to move this PR forward?

lib/src/inline_syntaxes/footnote_ref_syntax.dart
srawlins: Hmm no I don't think it needs to be renamed. Maybe it can just reference `[FootnoteDefSyntax]` so that there is a documentation link between the two. Then it's fine. Thanks!

lib/src/block_syntaxes/footnote_def_syntax.dart
srawlins: Sorry, I was on vacation for a few days 😭 but we're really almost there. I'd love to land this with like one more change 😁 . It is totally fine that there is not a published spec; that's all we have so that's ok. I do still love this feature because it helps us match what GitHub parses. So thanks again for your work!

With a feature that is under-specified like this, I think it' just important to add all of the links we can 😁 If you could add, in the doc comment links to:

* the issue: https://github.com/github/cmark-gfm/issues/283#issuecomment-1378868725
* the cmark-gfm source: https://github.com/github/cmark-gfm/blob/master/src/blocks.c#L1212
* the rust implementation is a great resource too, I think you can add it for good measure: https://github.com/wooorm/markdown-rs/blob/2498e31eecead798efc649502bbf5f86feaa94be/src/construct/gfm_footnote_definition.rs

srawlins (approved):
Thanks so much for all of the doc links. That really helps ❤️ ❤️ ❤️ 

I love it, thanks!

kevmoo:
We need to update the changelog and the version here, friends!

CC @lindeer @srawlins

srawlins:
Will do :)

Issue #445, opened 2022-08-18
Bug when parsing GFM extension for GitHub task lists (aka checkboxes).


isoos:
Markdown version: 6.0.0
Original report: https://github.com/dart-lang/pub-dev/issues/5963

Reproduction input that fails:
```md
- [x] Add nodeOnTap
- [] Animations
```

```
dart run markdown --extension-set GitHubWeb graphview.md

Unhandled exception:
RangeError: Value not in range: 8
#0      _RegExpMatch.group (dart:core-patch/regexp_patch.dart:166:7)
#1      _RegExpMatch.[] (dart:core-patch/regexp_patch.dart:178:17)
#2      ListSyntax.parse (package:markdown/src/block_syntaxes/list_syntax.dart:137:47)
#3      BlockParser.parseLines (package:markdown/src/block_parser.dart:123:32)
#4      Document.parseLines (package:markdown/src/document.dart:66:44)
#5      markdownToHtml (package:markdown/src/html_renderer.dart:42:26)
#6      main (file:///home/isoos/.pub-cache/hosted/pub.dartlang.org/markdown-6.0.0/bin/markdown.dart:57:11)
#7      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:295:32)
#8      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)
```

When changing `[]` (no space) to `[ ]` it works again.

srawlins:
Ouch, good catch.

srawlins:
I've got a fix in progress.

PR #446, opened 2022-08-24
Wrote some crashtest for running against markdown files from pub.dev

jonasfj:
This might be total overkill :rofl: 

I found 26 cases where it crashed all instances of https://github.com/dart-lang/markdown/issues/445.

It took about 20 minutes to run the tests on my desktop. I'm not sure it's worthwhile merging / keeping it around.
But it could be useful to run once in a while, because there should be no text input that causes the markdown parser to throw.

### DIFF
--- a/dart_test.yaml
+++ b/dart_test.yaml
@@ -1,0 +1,6 @@
+tags:
+  crashtest:
+    skip: 'Only run crashtest tests manually with `dart test -P crashtest`'
+    presets:
+      crashtest:
+        skip: false # Don't skip when running in -P crashtest
--- a/dart_test.yaml
+++ b/dart_test.yaml
@@ -1,6 +1,6 @@
 tags:
-  crashtest:
-    skip: 'Only run crashtest tests manually with `dart test -P crashtest`'
+  crash_test:
+    skip: 'Only run crash_test tests manually with `dart test -P crash_test`'
     presets:
-      crashtest:
-        skip: false # Don't skip when running in -P crashtest
+      crash_test:
+        skip: false # Don't skip when running in -P crash_test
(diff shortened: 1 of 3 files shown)


test/crash_test.dart
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
srawlins, line 1: 2022

test/crash_test.dart
```diff
@@ -0,0 +1,122 @@
...
+            skipped++;
+            return;
+          } on IOException {
+            skipped++;
+            return;
+          }
+          try {
+            await TarReader.forEach(Stream.value(archive), (entry) async {
+              if (entry.name.endsWith('.md')) {
+                late String str;
```
srawlins, line 82: Maybe a more meaningful name like 'text' or 'contents'.

dart_test.yaml
```diff
@@ -0,0 +1,6 @@
+tags:
+  crashtest:
```
srawlins, line 2: Should this be two words, `crash_test` to be consistent?

srawlins (approved):
Super cool, I think we should keep this around!

kevmoo:
@jonasfj – still running on this?

pubspec.yaml
```diff
@@ -21,9 +21,12 @@ dev_dependencies:
   build_web_compilers: ^3.0.0
   collection: ^1.15.0
   html: ^0.15.0
+  http: ^0.13.5
   io: ^1.0.0
   js: ^0.6.3
   lints: ^2.0.0
   path: ^1.8.0
+  pool: ^1.5.1
+  tar: ^0.5.5+1
```
devoncarew, line 30: fyi, this package has not been imported into google3

PR #447, opened 2022-08-26
Add an `enableTagfilter` option to `HtmlRenderer` to eanble GFM `tagfilter` extension

chenzhiguang:
This is a GFM extension, see: https://github.github.com/gfm/#disallowed-raw-html-extension-

But it still fails the test case. it is because of another issue.

### DIFF
--- a/lib/src/html_renderer.dart
+++ b/lib/src/html_renderer.dart
@@ -20,6 +20,7 @@ String markdownToHtml(
   Resolver? imageLinkResolver,
   bool inlineOnly = false,
   bool encodeHtml = true,
+  bool enableTagfilter = false,
   bool withDefaultBlockSyntaxes = true,
   bool withDefaultInlineSyntaxes = true,
 }) {
@@ -41,11 +42,14 @@ String markdownToHtml(
 
   final nodes = document.parseLines(lines);
 
-  return '${renderToHtml(nodes)}\n';
+  return '${renderToHtml(nodes, enableTagfilter: enableTagfilter)}\n';
 }
 
 /// Renders [nodes] to HTML.
-String renderToHtml(List<Node> nodes) => HtmlRenderer().render(nodes);
+String renderToHtml(List<Node> nodes, {bool enableTagfilter = false}) =>
+    HtmlRenderer(
+      enableTagfilter: enableTagfilter,
+    ).render(nodes);
 
 const _blockTags = [
   'blockquote',
@@ -92,8 +96,11 @@ class HtmlRenderer implements NodeVisitor {
 
   final _elementStack = <Element>[];
   String? _lastVisitedTag;
+  final bool _tagfilterEnabled;
 
-  HtmlRenderer();
+  HtmlRenderer({
+    bool enableTagfilter = false,
+  }) : _tagfilterEnabled = enableTagfilter;
 
   String render(List<Node> nodes) {
     buffer = StringBuffer();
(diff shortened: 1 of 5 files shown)


srawlins:
Can you explain why this should be an option on `HtmlRenderer`, rather than an extension added to an extension set?

tool/stats_lib.dart
```diff
@@ -118,10 +123,21 @@ CompareResult compareResult(
   bool verboseFail = false,
   bool verboseLooseMatch = false,
 }) {
+  var enableTagfilter = false;
+
+  for (final extension in testCase.extensions ?? []) {
```
srawlins, line 128: This can just use `testCase.extensions.contains`, yes?

tool/stats_lib.dart
```diff
@@ -94,6 +96,9 @@ class CommonMarkTestCase {
       json['end_line'] as int,
       json['markdown'] as String /*!*/,
       json['html'] as String,
+      json['extensions'] == null
```
srawlins, line 99: This is a lot of overhead and special casing for this one extension. It would be better if this could all be done like the other extensions.

lib/src/html_renderer.dart
```diff
@@ -104,6 +111,11 @@ class HtmlRenderer implements NodeVisitor {
   @override
   void visitText(Text text) {
     var content = text.textContent;
+
+    if (_tagfilterEnabled) {
```
srawlins, line 115: It seems this should be implemented in the parsers. It adds complexity to do anything but the most simple translation of the HTML nodes here to text. (I'm not crazy about the special line-splitting logic below, for example...)

srawlins (changes requested):

chenzhiguang:
> Can you explain why this should be an option on `HtmlRenderer`, rather than an extension added to an extension set?

Thank you for reviewing.

If use extensionSet to enable this extension, we need to create two syntaxes, one to parse block html, another to parse inline html. My solution is easier.

tool/stats_lib.dart
chenzhiguang: Yes, it is. I will change to `contains`. 

I was thinking of handling other extensions, but `testCase.extensions.contains` also works. maybe I was thinking there could be some extension combination which `contains` could not handle.

chenzhiguang:
I will try to create the syntax extensions and implement them with extensionSet later.

chenzhiguang:
I found the problem. I have tried the syntax solution first already, but I just forgot the big reason why I did not adopt it. for example:

Let's assume we have created `BlockTagfilterSyntax` and `InlineTagfilterSyntax`, we expect

```html
<blockquote>
  <xmp> is disallowed.  <XMP> is also disallowed.
</blockquote>
```
to be parsed to 
```html
<blockquote>
  &lt;xmp> is disallowed.  &lt;XMP> is also disallowed.
</blockquote>
```

But `<blockquote>` will be captured by `OtherTagBlockHtmlSyntax` until the  end of the closing tag `</blockquote>`
(`BlockTagfilterSyntax` would not capture `<blockquote>` because this tag is a tag should be filtered)

`OtherTagBlockHtmlSyntax` and all the other block html syntaxes just simply output the  original tags and the raw text between the opening and closing tags. So `InlineTagfilterSyntax` would not have a chance to be called, the actual output will be the same as the input:

```html
<blockquote>
  <xmp> is disallowed.  <XMP> is also disallowed.
</blockquote>
```

kevmoo:
Now that we've decided to move to a breaking change, maybe we can land this?

chenzhiguang:
I will rebase this PR when #471 is merged.

kevmoo:
We should also have a changelog entry for this, right?

chenzhiguang:
> We should also have a changelog entry for this, right?

Yes. Fixed

lib/src/html_renderer.dart
srawlins: OK I think this makes sense.

srawlins (approved):

Issue #448, opened 2022-08-26
{Ordered,Unordered}ListWithCheckBoxSyntax should allow lists where the first item is not a checkbox


srawlins:
GitHub is the spec here, and on GitHub, the following generates a checkbox list, with some items getting checkboxes, even though the first does not have a checkbox:

```markdown
* hello
* [ ] there
```

renders as:

```html
<ul class="contains-task-list">
<li>hello</li>
<li class="task-list-item"><input type="checkbox" id="" disabled="" class="task-list-item-checkbox"> there</li>
</ul>
```

The markdown package currently does _not_ parse such input as a list-with-checkboxes.

chenzhiguang:
https://github.github.com/gfm/#task-list-items-extension- says:

> GFM enables the tasklist extension, where an additional processing step is performed on [list items](https://github.github.com/gfm/#list-items).

Maybe we should try to add this additional process on list syntax instead? (parse the captured list items)

srawlins:
Could be. Whatever's easiest, and maintains the mechanism of using `{Ordered,Unordered}ListWithCheckBoxSyntax` in order to enable the syntax.

chenzhiguang:
@srawlins Here we go https://github.com/dart-lang/markdown/pull/450

chenzhiguang:
@srawlins This issue has been fixed already.

PR #449, opened 2022-08-26
Fix crash with mixed checkbox lists

srawlins:
Fixes https://github.com/dart-lang/markdown/issues/445

We have some more issues which I just raised in #448 , but this will fix the crashes.


keertip (approved):

isoos:
Do you have any estimates when this (and accompanied fixes) will get released?

srawlins:
I can try to release today.

PR #452, opened 2022-08-31
Return list for link nodes creation

lindeer:
DelimiterSyntax may create multiple nodes, and in `LinkSyntax`, nodes creation may need more params. @srawlins

### DIFF
--- a/lib/src/inline_parser.dart
+++ b/lib/src/inline_parser.dart
@@ -155,7 +155,7 @@ class InlineParser {
     final syntax = delimiter.syntax;
     if (syntax is LinkSyntax && syntaxes.any((e) => e is LinkSyntax)) {
       final nodeIndex = _tree.lastIndexWhere((n) => n == delimiter.node);
-      final linkNode = syntax.close(this, delimiter, null, getChildren: () {
+      final linkNodes = syntax.close(this, delimiter, null, getChildren: () {
         _processDelimiterRun(index);
         // All of the nodes which lie past [index] are children of this
         // link/image.
@@ -163,14 +163,14 @@ class InlineParser {
         _tree.removeRange(nodeIndex + 1, _tree.length);
         return children;
       });
-      if (linkNode != null) {
+      if (linkNodes != null) {
         _delimiterStack.removeAt(index);
         if (delimiter.char == $lbracket) {
           for (final d in _delimiterStack.sublist(0, index)) {
             if (d.char == $lbracket) d.isActive = false;
           }
         }
-        _tree.replaceRange(nodeIndex, _tree.length, linkNode);
+        _tree.replaceRange(nodeIndex, _tree.length, linkNodes);
         advanceBy(1);
         start = pos;
       } else {
@@ -240,7 +240,7 @@ class InlineParser {
         final openerTextNodeIndex = _tree.indexOf(openerTextNode);
         final closerTextNode = closer.node;
         var closerTextNodeIndex = _tree.indexOf(closerTextNode);
-        final node = opener.syntax.close(
+        final nodes = opener.syntax.close(
           this,
           opener,
           closer,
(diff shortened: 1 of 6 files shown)


lindeer:
@srawlins Footnote code would be like:
```
Iterable<Node> createFootnoteRef() {
  if (secondary) {
    result.add(Text('['));
    result.addAll(children);
    result.add(Text(']'));
  }
  ...
  final sup = Element('sup', [link])..attributes['class'] = 'footnote-ref';
  result.add(sup);
  return result;
}
```

srawlins:
Since this is a breaking change, we'll have to wait some time; flutter_markdown is still dealing with the 6.0.0 release.

lindeer:
OK, fine~

lindeer:
hi @srawlins , could you have a new review? seems flutter_markdown already updated markdown to 6.0.0

srawlins:
In 7 days, markdown 6.0.0 will be 60 days old, which I think is sufficient to not be too much "churn" for markdown clients.

lindeer:
60-days safety limit reached :grin: @srawlins

lindeer:
waiting for 2 months, is there any hope to merge?

srawlins:
There is hope. This PR is not the highest priority item, so it will not be today. Apologies.

lib/src/inline_syntaxes/delimiter_syntax.dart
```diff
@@ -83,15 +83,15 @@ class DelimiterSyntax extends InlineSyntax {
   ///
   /// If a tag can be closed at the current position, then this method calls
   /// [getChildren], in which [parser] parses any nested text into child nodes.
-  /// The returned [Node] incorpororates these child nodes.
-  Node? close(
+  /// The returned [Iterable] incorpororate these child nodes.
```
srawlins, line 86: Haha I misspelled this word to something crazy, let's use "includes"

lib/src/inline_parser.dart
```diff
@@ -170,7 +170,7 @@ class InlineParser {
             if (d.char == $lbracket) d.isActive = false;
           }
         }
-        _tree[nodeIndex] = linkNode;
+        _tree.replaceRange(nodeIndex, _tree.length, linkNode);
```
srawlins, line 173: I don't understand this bit. Does this change anything?

lib/src/inline_syntaxes/link_syntax.dart
```diff
@@ -147,18 +160,26 @@ class LinkSyntax extends DelimiterSyntax {
   }
 
   /// Tries to create a reference link node.
+  /// [secondary] indicates whether [label] is inside secondary `[]`
```
srawlins, line 163: I don't understand the `secondary` bit here. And it is unused, so can we instead add it in the follow-up Footnote PR?

srawlins (changes requested):
Looks pretty good! Could you bump the CHANGELOG as well, to 7.0.0-dev, with a note on the breaking API change? Thanks!

lib/src/inline_parser.dart
lindeer: now `linkNode` is a list, have to use list operation

srawlins: Oh I see; let's rename the variable, `linkNode` then.

lib/src/inline_parser.dart
```diff
@@ -255,7 +255,7 @@ class InlineParser {
         _tree.replaceRange(
           openerTextNodeIndex + 1,
           closerTextNodeIndex,
-          [node!],
+          node!,
```
srawlins, line 258: This `node` variable should be renamed as well, to reflect that its a List.

srawlins (changes requested):

srawlins (approved):
I think this looks great, thanks!

PR #456, opened 2022-09-08
lint cleanup

kevmoo:


analysis_options.yaml
```diff
@@ -20,10 +20,12 @@ linter:
     - directives_ordering
     - only_throw_errors
     - package_api_docs
+    - prefer_const_declarations
+    - prefer_final_locals
```
srawlins, line 24: Just slid that one in there, did ya?

srawlins: Oh, lol it's just sorted. I was like, "how is this CL not much much bigger???"

kevmoo: I did add `prefer_const_declarations' 😁

PR #458, opened 2022-09-15
Refactor HTML block syntax

chenzhiguang:
To make the code clean and easier to manage.


chenzhiguang:
Hi @srawlins! This PR might be a good one?

kevmoo (commented):
Great to see the standards bits creeping up! Leaving to @srawlins to review the code!

lib/src/patterns.dart
```diff
@@ -124,3 +124,73 @@ final tablePattern = RegExp(
...
+    r'\s*/?>'
+
+    //
+    '|'
+
+    // Closing tag, see
+    // https://spec.commonmark.org/0.30/#closing-tag.
+    r'</[a-z][a-z0-9-]*\s*>';
+
+final htmlBlockPattern = RegExp(
```
kevmoo, line 167: Could this be private? Is it referenced anywhere else?

chenzhiguang: Yes, it could be private for now, currently it is only used in `HtmlBlockSyntax`. The idea was to manage all these patterns at one file, also we do not export these patterns.

But it's dependency `_namedTagDefinition` pattern might have another consumer later, I am planing to improve the inline HTML syntax, it's pattern also relay on `_namedTagDefinition`.

kevmoo: If there are reasons to share bits between them, sure. If they are only used in one library, better to keep them in one library (and make them private )IMHO

lib/src/block_syntaxes/html_block_syntax.dart
```diff
@@ -0,0 +1,84 @@
...
+        lines.add(parser.current);
+        if (endCondition.hasMatch(parser.current)) {
+          break;
+        }
+        parser.advance();
+      }
+      parser.advance();
+    }
+
+    // If the following lines start a HTML block again, put them together with
```
srawlins, line 67: grammar nit: "an HTML"

lib/src/patterns.dart
```diff
@@ -124,3 +124,73 @@ final tablePattern = RegExp(
...
+    // https://spec.commonmark.org/0.30/#unquoted-attribute-value.
+    r'''(?:[^\s"'=<>`]+?|'[^']*?'|"[^"]*?")'''
+
+    // Attribute ends.
+    ')?)*'
+
+    // Opening tag ends.
+    r'\s*/?>'
+
+    //
```
srawlins, line 160: Is the comment here just to preserve newlines? Could say something like 'OR'

lib/src/patterns.dart
```diff
@@ -124,3 +124,73 @@ final tablePattern = RegExp(
...
+    '</?(?<condition_6>address|article|aside|base|basefont|blockquote|body|'
+    'caption|center|col|colgroup|dd|details|dialog|dir|DIV|dl|dt|fieldset|'
+    'figcaption|figure|footer|form|frame|frameset|h1|h2|h3|h4|h5|h6|head|'
+    'header|hr|html|iframe|legend|li|link|main|menu|menuitem|nav|noframes|ol|'
+    'optgroup|option|p|param|section|source|summary|table|tbody|td|tfoot|th|'
+    'thead|title|tr|track|ul)'
+    r'(?:\s|>|/>|$)'
+    '|'
+
+    // Here is more restricted than the commonmark definition(Rule #7).
```
srawlins, line 189: Grammar, I think something like "Here we are more restrictive than the Commonmark definition (Rule #7)."

lib/src/patterns.dart
```diff
@@ -124,3 +124,73 @@ final tablePattern = RegExp(
...
+    // Otherwise some raw HTML test cases will fail, for example:
+    // https://spec.commonmark.org/0.30/#example-618.
+    // Because if a line is treated as a HTML block, it will output as Text node
```
srawlins, line 192: grammar nit: "an HTML"

tool/common_mark_stats.json
```diff
@@ -333,7 +333,7 @@
   "168": "strict",
   "169": "strict",
   "170": "strict",
-  "171": "loose",
```
srawlins, line 336: 🙌

lib/src/patterns.dart
```diff
@@ -124,3 +124,73 @@ final tablePattern = RegExp(
...
+    //
+    '|'
+
+    // Closing tag, see
+    // https://spec.commonmark.org/0.30/#closing-tag.
+    r'</[a-z][a-z0-9-]*\s*>';
+
+final htmlBlockPattern = RegExp(
+    '^ {0,3}(?:'
+    '<(?<condition_1>pre|script|style|textarea)'
```
srawlins, line 169: Can you add a (probably huge, sorry) comment for `htmlBlockPattern` noting what the condition groups are?

srawlins (changes requested):
I love any PR which strictly improves our test coverage! Improving readability is a huge win too!

lib/src/block_syntaxes/html_block_syntax.dart
kevmoo:
```suggestion
    // If the following lines start an HTML block again, put them together with
```

lib/src/patterns.dart
chenzhiguang: Fixed

lib/src/patterns.dart
chenzhiguang: Fixed

lib/src/patterns.dart
chenzhiguang: Fixed!

tool/common_mark_stats.json
chenzhiguang: Thanks!

lib/src/patterns.dart
chenzhiguang: These 7 conditions correspond to the ones in the Commonmark specification, I have added a short comment and a link. Maybe it works?

lib/src/block_syntaxes/html_block_syntax.dart
```diff
@@ -0,0 +1,84 @@
...
+import '../block_parser.dart';
+import '../patterns.dart';
+import 'block_syntax.dart';
+
+/// Parse HTML blocks.
+class HtmlBlockSyntax extends BlockSyntax {
+  @override
+  RegExp get pattern => htmlBlockPattern;
+
+  // All types of HTML blocks except type 7 may interrupt a paragraph.
```
srawlins, line 15: This comment needs some context too. The "type 7" part. Maybe expand here, or link to commonmark, or add context to the comment above on line 10. Just because this is the first mention of "type 7" when reading the file from top to bottom.

lib/src/patterns.dart
srawlins: That's awesome.

srawlins (changes requested):
One more small request, and then I love this.

lib/src/block_syntaxes/html_block_syntax.dart
chenzhiguang: Done!

srawlins (approved):

PR #463, opened 2022-09-29
Use fewer empty lists; non-growable

srawlins:
Mainly this replaces a few `foo ?? []` cases with something that does not unnecessarily create a new list. Maybe simpler readability too.


kevmoo (approved):
:shipit:

PR #465, opened 2022-10-04
Fix unicode case folding

chenzhiguang:
This PR simply implements the **Unicode Case folding** on `normalizeLinkLabel` function.

See https://spec.commonmark.org/0.30/#link-label

> To normalize a label, strip off the opening and closing brackets, perform the **Unicode case fold**, strip leading and trailing spaces

This PR also includes a script to generate the `caseFoldingMap`.

Also I have created an `assets` folder,  my idea is to put these generated files together in this folder, now we have `emojis.dart` and `case_folding.dart`. There will be an `html_entities.dart` file later soon. If you agree with it, I will create another PR to move `emojis.dart` into it.


chenzhiguang:
Hi @srawlins! How about this one?

tool/common_mark_stats.txt
```diff
@@ -24,5 +24,5 @@
   11 of   11 – 100.0%  Tabs
    3 of    3 – 100.0%  Textual content
   19 of   19 – 100.0%  Thematic breaks
- 620 of  652 –  95.1%  TOTAL
- 553 of  620 –  89.2%  TOTAL Strict
+ 621 of  652 –  95.2%  TOTAL
+ 554 of  621 –  89.2%  TOTAL Strict
```
kevmoo, line 28: This makes me so happy! 😀

chenzhiguang: Thanks, me too. The next ones could be better. 😄

kevmoo: You've got us +7 with your commits in the last few weeks!

lib/src/util.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 
+import '../assets/case_folding.dart';
```
kevmoo, line 7: Perhaps `src/assets/...` – all things that should not be imported by users should be in `src`

kevmoo: 👍

kevmoo (approved):

kevmoo:
LGTM, but @srawlins should also give his blessing!

lib/src/util.dart
```diff
@@ -83,9 +84,15 @@ final _oneOrMoreWhitespacePattern = RegExp('[ \n\r\t]+');
...
 /// "Normalizes" a link label, according to the [CommonMark spec].
 ///
-/// [CommonMark spec] https://spec.commonmark.org/0.29/#link-label
-String normalizeLinkLabel(String label) =>
-    label.trim().replaceAll(_oneOrMoreWhitespacePattern, ' ').toLowerCase();
+/// [CommonMark spec] https://spec.commonmark.org/0.30/#link-label
+String normalizeLinkLabel(String label) {
+  final text = label.trim().replaceAll(_oneOrMoreWhitespacePattern, ' ');
+  final buffer = StringBuffer();
+  for (var i = 0; i < text.length; i++) {
```
srawlins, line 91: My only concern with this PR is performance. At this line. Two bits:

* For every character, we  have to do these two lookups (`text[]` and `caseFoldingMap[]`)
* Even if `text[i]` is not in `caseFoldingMap`, we still perform an action (add to the StringBuffer).

Can you post some benchmark results from `benchmark/benchmark.dart`, please?

srawlins (changes requested):

lib/src/util.dart
chenzhiguang: ### Now
Screenshot 2022-10-06 at 20 36 16

### Before
Screenshot 2022-10-06 at 20 36 55

chenzhiguang:
Screenshot 2022-10-06 at 21 20 08

Now it is slightly better

srawlins:
Haha it seems nearly 0 performance change. 🎉

lib/src/util.dart
```diff
@@ -83,9 +84,17 @@ final _oneOrMoreWhitespacePattern = RegExp('[ \n\r\t]+');
...
-/// [CommonMark spec] https://spec.commonmark.org/0.29/#link-label
-String normalizeLinkLabel(String label) =>
-    label.trim().replaceAll(_oneOrMoreWhitespacePattern, ' ').toLowerCase();
+/// [CommonMark spec] https://spec.commonmark.org/0.30/#link-label
+String normalizeLinkLabel(String label) {
+  var text = label.trim().replaceAll(_oneOrMoreWhitespacePattern, ' ');
+  for (var i = 0; i < text.length; i++) {
+    final mapped = caseFoldingMap[text[i]];
+    if (mapped != null) {
+      text = text.replaceRange(i, i + 1, mapped);
```
srawlins, line 93: It's unfortunate that this is a little verbose, and `replaceRange` sounds like it can't be too performant, haha. Though maybe the backends have optimizations for ranges of size 1.

In any case, all of this code is only run for link labels, so, 🎉 , it's great.

srawlins (approved):

PR #467, opened 2022-10-05
Fix html entity and numeric character references

chenzhiguang:
This PR fixes the HTML entity references and numeric character references issues, introduces a new syntax `DecodeHtmlSyntax`.

Unfortunately this PR looses one **strict**.  It should be fixed in the later HTML escape improvement PR soon.


chenzhiguang:
Hello @srawlins, here is another one. There are more PRs on the way 🙂

chenzhiguang:
benchmarks:

### Now
Screenshot 2022-10-06 at 20 45 22

### Before
Screenshot 2022-10-06 at 20 36 55

srawlins:
Hmm so this one adds ~5% to the time, which is not super small...

chenzhiguang:
Yes, it does have a big drop. Maybe it is hard to avoid it because a new extra `InlineSyntax` has been implemented.

This PR is a part of my **escape and html characters improvement** epic. The good news is after I have strictly improved the test coverage for 30 more, the performance will come back a little bit:

### Benchmark
Screenshot 2022-10-07 at 04 46 25

### Test coverage
Screenshot 2022-10-07 at 05 43 47

Maybe I could have a performance session after that.

kevmoo (commented):
I love that we're catching up with the standard here.

Maybe it'd be worth seeing if another output could be faster here. Like generating a switch statement or something?

chenzhiguang:
> I love that we're catching up with the standard here.
> 
> Maybe it'd be worth seeing if another output could be faster here. Like generating a switch statement or something?

Thanks, it think it could improve the performance a bit.

Now the drop comes from the syntax match mostly. You can see it from this screen record and the screenshot below.

https://www.dropbox.com/s/digj1fhg4cur05m/markdown-decode-html-syntax.mov?dl=0
 
Screenshot 2022-10-14 at 05 16 53

kevmoo:
"worst" perf is 1.18ms – to 1.12ms – so the delta is 0.06 MILLIseconds?

chenzhiguang:
> "worst" perf is 1.18ms – to 1.12ms – so the delta is 0.06 MILLIseconds?

Yes, I think so.

kevmoo:
Thoughts @srawlins ?

kevmoo:
@chenzhiguang – I added some lints. You might want to rebase this on master!

chenzhiguang:
> @chenzhiguang – I added some lints. You might want to rebase this on master!

Done!

kevmoo:
@srawlins – thoughts here?

chenzhiguang:
@srawlins Hi, do we have a plan for this PR, as it blocks my progress, my plan is to fix all these decoding, encoding, escaping issues. Decoding is the first step.

lib/src/inline_syntaxes/decode_html_syntax.dart
```diff
@@ -0,0 +1,48 @@
...
+
+  @override
+  bool tryMatch(InlineParser parser, [int? startMatchPos]) {
+    if (parser.pos > 0 && parser.charAt(parser.pos - 1) == $backquote) {
+      return false;
+    }
+    final match = pattern.matchAsPrefix(parser.source, parser.pos);
+    if (match == null ||
+        match[1] != null ||
+        (match[2] == null && match[3] == null)) {
```
srawlins, line 28: This is rather inscrutible, haha. Would you mind adding a comment, or creating local variables with meaningful names (e.g. for `match[1]` which is recalled on line 32).

tool/update_entities.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+import 'package:path/path.dart' as p;
+
+// Generates and updates HTML entities.
```
srawlins, line 9: This can be a doc comment (`///`).

srawlins (changes requested):
I love it. Just a few requests.

lib/src/inline_syntaxes/decode_html_syntax.dart
```diff
@@ -22,14 +21,16 @@ class DecodeHtmlSyntax extends InlineSyntax {
...
-        (match[2] == null && match[3] == null)) {
+    if (match == null) {
       return false;
     }
 
-    if (match[1] != null && htmlEntitiesMap[match.match] == null) {
+    // TODO(Zhiguang): Enable HTML entity decoding when working on HTML escape
+    // issues.
+    final isEntity = match[1] != null;
+    if (isEntity) {
```
kevmoo, line 33: Eh...don't think this helps. Just putting the `match[1] != null` in the `if` block is cleaner

lib/src/inline_syntaxes/decode_html_syntax.dart
chenzhiguang: Yep, it is confusing. 

I have simplified the `if` condition. 

I guess I was trying to temporarily disable HTML entity decoding but keep most of the logic, so I can  easily enable it again later when dealing with HTML escaping issues.

```dart
match[1] != null;  // it  is an HTML entity reference
match[2] != null;  // it is a decimal numeric character references 
match[3] != null;  // it is a hexadecimal numeric character references
```

tool/update_entities.dart
chenzhiguang: Fixed

lib/src/inline_syntaxes/decode_html_syntax.dart
chenzhiguang: Yep, fixed.

I will give them meaningful names later when it becomes confusing. :)

lib/src/util.dart
```diff
@@ -96,6 +98,49 @@ String normalizeLinkLabel(String label) {
...
+
+  // Entity references, see
+  // https://spec.commonmark.org/0.30/#entity-references.
+  if (match[1] != null) {
+    decodedText = htmlEntitiesMap[text];
+  }
+
+  // Decimal numeric character references, see
+  // https://spec.commonmark.org/0.30/#decimal-numeric-character-references.
+  else if (match[2] != null) {
```
srawlins, line 119: This would be better as a local variable, to avoid the second lookup and null-check.

lib/src/util.dart
```diff
@@ -96,6 +98,49 @@ String normalizeLinkLabel(String label) {
...
+      hexValue = 0xFFFd;
+    }
+
+    decodedText = String.fromCharCode(hexValue);
+  }
+
+  // Hexadecimal numeric character references, see
+  // https://spec.commonmark.org/0.30/#hexadecimal-numeric-character-references.
+  else if (match[3] != null) {
+    var hexValue = int.parse(match[3]!, radix: 16);
```
srawlins, line 134: Same here, making `match[3]` a local variable will avoid the second look up and  null check.

srawlins (changes requested):

lib/src/util.dart
chenzhiguang: Agreed, fixed!

lib/src/util.dart
chenzhiguang: Fixed

lib/src/util.dart
```diff
@@ -96,6 +98,52 @@ String normalizeLinkLabel(String label) {
...
+  // Entity references, see
+  // https://spec.commonmark.org/0.30/#entity-references.
+  if (isEntity) {
+    decodedText = htmlEntitiesMap[text];
+  }
+
+  // Decimal numeric character references, see
+  // https://spec.commonmark.org/0.30/#decimal-numeric-character-references.
+  else if (isDecimal) {
+    final decimalValue = int.parse(match[2]!);
```
srawlins, line 123: The map access and null-check are still repeated here. We get `match[2]` above, and check if its null, then we might enter here and get `match[2]` again, and check if it is null, again.

It would be better to assign `match[2]` to a local variable and allow one null-check to promote.

lib/src/util.dart
```diff
@@ -96,6 +98,52 @@ String normalizeLinkLabel(String label) {
...
+      hexValue = 0xFFFd;
+    }
+
+    decodedText = String.fromCharCode(hexValue);
+  }
+
+  // Hexadecimal numeric character references, see
+  // https://spec.commonmark.org/0.30/#hexadecimal-numeric-character-references.
+  else if (isHexadecimal) {
+    var hexValue = int.parse(match[3]!, radix: 16);
```
srawlins, line 137: Same with `match[3]!`; assigning to a local variable will simplify this.

lib/src/util.dart
```diff
@@ -96,6 +98,52 @@ String normalizeLinkLabel(String label) {
...
+  else if (isDecimal) {
+    final decimalValue = int.parse(match[2]!);
+    int hexValue;
+    if (decimalValue < 1114112 && decimalValue > 1) {
+      hexValue = int.parse(decimalValue.toRadixString(16), radix: 16);
+    } else {
+      hexValue = 0xFFFd;
+    }
+
+    decodedText = String.fromCharCode(hexValue);
```
srawlins, line 131: I think it would be simpler to just return this value. `return String.fromCharCode(hexValue);`.

lib/src/util.dart
```diff
@@ -96,6 +98,52 @@ String normalizeLinkLabel(String label) {
...
+  }
+
+  // Hexadecimal numeric character references, see
+  // https://spec.commonmark.org/0.30/#hexadecimal-numeric-character-references.
+  else if (isHexadecimal) {
+    var hexValue = int.parse(match[3]!, radix: 16);
+    if (hexValue > 0x10ffff || hexValue == 0) {
+      hexValue = 0xFFFd;
+    }
+    decodedText = String.fromCharCode(hexValue);
```
srawlins, line 141: Here too, we can just return instead of using this local variable.

lib/src/util.dart
```diff
@@ -96,6 +98,52 @@ String normalizeLinkLabel(String label) {
...
+  final text = match.match;
+  final isEntity = match[1] != null;
+  final isDecimal = match[2] != null;
+  final isHexadecimal = match[3] != null;
+  String? decodedText;
+
+  // Entity references, see
+  // https://spec.commonmark.org/0.30/#entity-references.
+  if (isEntity) {
+    decodedText = htmlEntitiesMap[text];
```
srawlins, line 117: Here we can return `htmlEntitiesMap[text] ?? text`.

srawlins (changes requested):

lib/src/util.dart
chenzhiguang: Thanks a lot! 

Fixed

lib/src/util.dart
chenzhiguang: Fixed

lib/src/util.dart
chenzhiguang: Fixed

lib/src/util.dart
chenzhiguang: Fixed

lib/src/util.dart

srawlins (approved):
Thanks much! Love the performance benchmarking and the new spec compliance!

chenzhiguang:
Thanks! I have leant a lot from it.

lib/src/inline_parser.dart
```diff
@@ -92,6 +93,7 @@ class InlineParser {
     if (document.withDefaultInlineSyntaxes) {
       // Custom link resolvers go after the generic text syntax.
       syntaxes.addAll([
+        DecodeHtmlSyntax(),
```
oprypin, line 96: The addition to the changelog says

> * Add a new syntax `DecodeHtmlSyntax` to decode HTML entity and numeric
  character references.

However, this changes the default as well, it's definitely an incompatible change. Was that intended?

And, how to opt out of this change, if necessary?

chenzhiguang: If you need to run the parser without `DecodeHtmlSyntax`, you have to set `withDefaultInlineSyntaxes` as `false` and add all the inline syntaxes you want to have through `inlineSyntaxes` option,  unfortunately there is not a better way at this moment. Maybe we could add options to make it possible to disable each individual syntax in the future.

oprypin: The reason that I'm inquiring is that this change makes the following test fail:

https://github.com/flutter/packages/blob/8496fe847adfdf64d0779dc683d2bbb93e960ba8/packages/flutter_markdown/test/html_test.dart#L56

and it is hard for me to judge whether that test's expectation is important for their purposes or not.

kevmoo: Also, flutter_markdown may need to bend towards the spec. We have users writing markdown that shows up on GitHub readme rendering. We should be mindful that we can't have bespoke one-off interpretations of the spec – it'll just annoy our users.

devoncarew: @stuartmorgan - who can speak to the package:flutter_markdown behavior / impl? package:markdown's impl is changing so that the flutter_markdown test `doesn't convert existing HTML entities when parsing` will no longer be true.

(we can also track this as in issue in flutter/flutter if that's easier for the conversation)

stuartmorgan: I'm not very familiar with the details of `flutter_markdown` (/cc @domesticmouse who does most of the stewardship for it), but its documented behavior is to follow the relevant specs, so if the test is going against the spec I would think we should just fix the test.

domesticmouse: I'm owner in as much as I make sure changes have test coverage. I don't have any feelings about compliance against the various flavours of Markdown standards (CommonMark, GitHub flavoured, et al). 

I'm happy for tests to be fixed in the direction of standards compliance...

devoncarew: I filed this to track removing the test: https://github.com/flutter/flutter/issues/114193 (and https://github.com/dart-lang/markdown/issues/477 for some additional pre-submit checking for package:markdown).

devoncarew: https://github.com/flutter/packages/pull/2757

PR #468, opened 2022-10-13
Misc package cleanup, mostly lints

kevmoo:
Centralized HTTP JSON download logic

### DIFF
--- a/benchmark/benchmark.dart
+++ b/benchmark/benchmark.dart
@@ -38,7 +38,8 @@ void main() {
     // the VM doesn't optimize "dead" code away.
     if (result != expected) {
       print('Incorrect output:\n$result');
-      exit(1);
+      exitCode = 1;
+      return;
     }
 
     // Don't print the first run. It's always terrible since the VM hasn't
--- a/test/util.dart
+++ b/test/util.dart
@@ -28,8 +28,8 @@ void testFile(
   Iterable<BlockSyntax> blockSyntaxes = const [],
   Iterable<InlineSyntax> inlineSyntaxes = const [],
 }) {
-  final directory = p.join(p.current, 'test');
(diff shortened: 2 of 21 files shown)


tool/update_shared.dart
```diff
@@ -0,0 +1,21 @@
...
+  try {
+    final request = await client.getUrl(Uri.parse(uri));
+    final response = await request.close();
+
+    return response
+        .transform(utf8.decoder)
+        .transform(const JsonDecoder())
+        .single;
+  } finally {
+    client.close();
```
kevmoo, line 19: Seeing the same bug three times put me over the edge! Also it's nice to use our stream API

benchmark/benchmark.dart
```diff
@@ -38,7 +38,8 @@ void main() {
     // the VM doesn't optimize "dead" code away.
     if (result != expected) {
       print('Incorrect output:\n$result');
-      exit(1);
+      exitCode = 1;
```
kevmoo, line 41: `exit(...)` makes debugging impossible. Cleaned these up, too!

test/util.dart
```diff
@@ -24,19 +23,13 @@ Future<void> testDirectory(String name, {ExtensionSet? extensionSet}) async {
   }
 }
 
-Future<String> get markdownPackageRoot async {
```
kevmoo, line 27: We can just assume tests are run from the root of the package. This vestigial logic is very much no longer needed.

devoncarew (approved):

test/util.dart
devoncarew: This may be causing test failures for the roll to google3.

We should revert this (or filter out related tests in google3).

kevmoo: Oh. Ugh.

Issue #470, opened 2022-10-24
Link creation bugs in GitHubWeb-flavored extension


isoos:
Original report on `pub-dev`: https://github.com/dart-lang/pub-dev/issues/6126
Markdown version: `6.0.1`

**1. Extra text content after the link closing tag:**
```
dart run markdown --extension-set GitHubWeb
(https://github.com/intel/rohd/issues/129)
<p><a href="https://github.com/intel/rohd/issues/129">https://github.com/intel/rohd/issues/129</a>9)</p>
```

**2. <a> href URL with extra parenthesis, also in the link text content:**
```
dart run markdown --extension-set GitHubWeb
(https://github.com/intel/rohd/issues/129).
<p><a href="https://github.com/intel/rohd/issues/129)">https://github.com/intel/rohd/issues/129)</a>).</p>
```

chenzhiguang:
Fixed.

isoos:
@chenzhiguang @srawlins: any ETA when this (and other fixes) will get released?

srawlins:
I'm hoping within a month.

PR #471, opened 2022-10-24
Refactor AutolinkExtensionSyntax

chenzhiguang:
To improve the readability and test coverage.

Fix: https://github.com/dart-lang/markdown/issues/470

### DIFF
--- a/lib/src/inline_syntaxes/autolink_extension_syntax.dart
+++ b/lib/src/inline_syntaxes/autolink_extension_syntax.dart
@@ -63,19 +63,18 @@ class AutolinkExtensionSyntax extends InlineSyntax {
   bool onMatch(InlineParser parser, Match match) {
     int consumeLength;
 
-    var isEmail = false;
-    if (match[2] == null) {
-      consumeLength = _getConsumeLength(match.match);
-    } else {
+    final isEmailLink = match[2] != null;
+    if (isEmailLink) {
       consumeLength = match.match.length;
-      isEmail = true;
+    } else {
+      consumeLength = _getConsumeLength(match.match);
     }
 
     var text = match.match.substring(0, consumeLength);
     text = parser.encodeHtml ? escapeHtml(text) : text;
 
     var destination = text;
-    if (isEmail) {
+    if (isEmailLink) {
       destination = 'mailto:$destination';
     } else if (!destination.startsWith(RegExp(r'(https?|ftp):\/\/'))) {
       destination = 'http://$destination';
--- a/lib/src/inline_syntaxes/autolink_extension_syntax.dart
+++ b/lib/src/inline_syntaxes/autolink_extension_syntax.dart
@@ -76,7 +76,8 @@ class AutolinkExtensionSyntax extends InlineSyntax {
     var destination = text;
     if (isEmailLink) {
       destination = 'mailto:$destination';
-    } else if (!destination.startsWith(RegExp(r'(https?|ftp):\/\/'))) {
+    } else if (destination[0] == 'w') {
+      // When there is no scheme specified, insert the scheme `http`.
       destination = 'http://$destination';
     }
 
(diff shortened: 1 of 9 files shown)


tool/stats_lib.dart
```diff
@@ -54,19 +61,16 @@ class Config {
   static final Config commonMarkConfig = Config._(
     'common_mark',
     'http://spec.commonmark.org/0.28/',
-    null,
   );
   static final Config gfmConfig = Config._(
     'gfm',
     'https://github.github.com/gfm/',
-    ExtensionSet.gitHubFlavored,
```
chenzhiguang, line 62: The `ExtensionSet.gitHubFlavored` does not work out here. Because `AutolinkExtensionSyntax` and `AutolinkSyntax` conflict to each other.

For example: 
https://github.github.com/gfm/#example-620
```markdown
http://example.com
```

Without `AutolinkExtensionSyntax`, the result is:
```html
<p>http://example.com</p>
```
With `AutolinkExtensionSyntax`, the result is:
```html
<p><a href="http://example.com">http://example.com</a></p>
```

chenzhiguang:
@srawlins hi, I have created 3 more PRs. I am not sure if I need to ping you each time when I have created a new PR, so it will show up somewhere on your dashboard. 🙂

tool/stats_lib.dart
```diff
@@ -117,11 +126,31 @@ CompareResult compareResult(
   bool throwOnError = false,
   bool verboseFail = false,
   bool verboseLooseMatch = false,
+  List<String> extensions = const [],
```
kevmoo, line 129: Should probably be a `Set<String>`

tool/stats_lib.dart
```diff
@@ -117,11 +126,31 @@ CompareResult compareResult(
...
+  for (final extension in extensions) {
+    switch (extension) {
+      case 'autolink':
+        inlineSyntaxes.add(AutolinkExtensionSyntax());
+        break;
+      case 'strikethrough':
+        inlineSyntaxes.add(StrikethroughSyntax());
+        break;
+      case 'table':
+        blockSyntaxes.add(const TableSyntax());
```
kevmoo, line 144: probably good to add a `default` case and throw if we hit unexpected input, right?

tool/stats_lib.dart
chenzhiguang: Fixed

tool/stats_lib.dart
chenzhiguang: Fixed

srawlins (changes requested):
Thanks! One question

lib/src/inline_syntaxes/autolink_extension_syntax.dart
```diff
@@ -3,134 +3,127 @@
...
-        final entityRefLength = entityRef.match.length;
-        url = url.substring(0, url.length - entityRefLength);
-        href = href.substring(0, href.length - entityRefLength);
-        matchLength -= entityRefLength;
-      }
+    var destination = text;
+    if (isEmailLink) {
+      destination = 'mailto:$destination';
+    } else if (!destination.startsWith(RegExp(r'(https?|ftp):\/\/'))) {
+      destination = 'http://$destination';
```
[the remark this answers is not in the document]
chenzhiguang: It happens when the scheme is missing such as `www.commonmark.org`, we insert an `http` automatically.

The spec does not mention anything about `ftp`, maybe it was removed from the new version, I am not sure if we should remove it as well.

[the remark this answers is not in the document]
chenzhiguang: Just did an update.

If a text is recognised by Autolinks Extension as a link(not an email) and it does not start with a URI scheme, it must start with 'www.', so we only need to check the first letter of this matched text.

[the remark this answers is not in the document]
srawlins: sounds good!

srawlins (approved):
Thanks!

PR #472, opened 2022-10-25
Optimise TableSyntax

chenzhiguang:
Now it is 100% conform to GFM table spec.

### DIFF
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -121,7 +121,7 @@ class BlockParser {
 
     while (!isDone) {
       for (final syntax in blockSyntaxes) {
-        if (neverMatch != null && neverMatch == syntax) {
+        if (neverMatch == syntax) {
           continue;
         }
 
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -112,13 +112,13 @@ class BlockParser {
   List<Node> parseLines() {
     final blocks = <Node>[];
 
-    // If the `_pos` does not change before and after parse(), never try to
-    // match this `_pos` with the same syntax again.
-    final neverMatch = <BlockSyntax>[];
+    // If the `_pos` does not change before and after `parse()`, never try to
+    // parse this `_pos` with the same syntax again.
+    BlockSyntax? neverMatch;
 
     while (!isDone) {
       for (final syntax in blockSyntaxes) {
-        if (neverMatch.contains(syntax)) {
+        if (neverMatch != null && neverMatch == syntax) {
           continue;
         }
 
@@ -128,12 +128,8 @@ class BlockParser {
           if (block != null) {
             blocks.add(block);
           }
+          neverMatch = _pos != positionBefore ? null : syntax;
 
-          if (_pos == positionBefore) {
-            neverMatch.add(syntax);
-          } else {
-            neverMatch.clear();
-          }
           break;
         }
       }
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -113,7 +113,7 @@ class BlockParser {
     final blocks = <Node>[];
 
     // If the `_pos` does not change before and after parse(), never try to
-    // match this `_pos` again.
+    // match this `_pos` with the same syntax again.
     final neverMatch = <BlockSyntax>[];
 
     while (!isDone) {
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -91,6 +91,10 @@ class BlockParser {
     _pos++;
   }
 
+  void retreat() {
+    _pos--;
+  }
+
   bool get isDone => _pos >= lines.length;
 
   /// Gets whether or not the current line matches the given pattern.
@@ -107,11 +111,29 @@ class BlockParser {
 
   List<Node> parseLines() {
     final blocks = <Node>[];
+
+    // If the `_pos` does not change before and after parse(), never try to
+    // match this `_pos` again.
+    final neverMatch = <BlockSyntax>[];
+
     while (!isDone) {
       for (final syntax in blockSyntaxes) {
+        if (neverMatch.contains(syntax)) {
+          continue;
+        }
+
         if (syntax.canParse(this)) {
+          final positionBefore = _pos;
           final block = syntax.parse(this);
-          if (block != null) blocks.add(block);
+          if (block != null) {
+            blocks.add(block);
+          }
+
+          if (_pos == positionBefore) {
+            neverMatch.add(syntax);
+          } else {
+            neverMatch.clear();
+          }
           break;
         }
       }
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -112,13 +112,13 @@ class BlockParser {
   List<Node> parseLines() {
     final blocks = <Node>[];
 
-    // If the `_pos` does not change before and after `parse()`, never try to
-    // parse this `_pos` with the same syntax again.
-    BlockSyntax? neverMatch;
+    // If the `_pos` does not change before and after parse(), never try to
+    // match this `_pos` again.
+    final neverMatch = <BlockSyntax>[];
 
     while (!isDone) {
       for (final syntax in blockSyntaxes) {
-        if (neverMatch != null && neverMatch == syntax) {
+        if (neverMatch.contains(syntax)) {
           continue;
         }
 
@@ -128,8 +128,12 @@ class BlockParser {
           if (block != null) {
             blocks.add(block);
           }
-          neverMatch = _pos != positionBefore ? null : syntax;
 
+          if (_pos == positionBefore) {
+            neverMatch.add(syntax);
+          } else {
+            neverMatch.clear();
+          }
           break;
         }
       }
--- a/lib/src/block_parser.dart
+++ b/lib/src/block_parser.dart
@@ -113,7 +113,10 @@ class BlockParser {
     final blocks = <Node>[];
 
     // If the `_pos` does not change before and after `parse()`, never try to
-    // parse this `_pos` with the same syntax again.
+    // parse the line at `_pos` with the same syntax again.
+    // For example the `TableSyntax` might not advance the `_pos` in `parse`
+    // method, beause of the header row does not match the delimiter row in the
+    // number of cells, which makes a table like structure not be recognized.
     BlockSyntax? neverMatch;
 
     while (!isDone) {
--- a/lib/src/block_syntaxes/table_syntax.dart
+++ b/lib/src/block_syntaxes/table_syntax.dart
@@ -83,7 +83,7 @@ class TableSyntax extends BlockSyntax {
       if (matchLeft) return 'left';
       if (matchRight) return 'right';
       return null;
-    }).toList(growable: false);
+    }).toList();
   }
 
   /// Parses a table row at the current line into a table row element, with
--- a/lib/src/block_syntaxes/table_syntax.dart
+++ b/lib/src/block_syntaxes/table_syntax.dart
@@ -72,16 +72,32 @@ class TableSyntax extends BlockSyntax {
   }
 
   List<String?> _parseAlignments(String line) {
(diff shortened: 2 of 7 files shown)


srawlins:
How does this contain commits which are already in dart-lang:master? I can't review a PR with commits like this... but I'm honestly not even sure how it's happened. CC @kevmoo, my git expert. What is needed here? A rebase?

chenzhiguang:
I did a rebase, then it messed up.

kevmoo:
Yeah, you want to do a `git rebase -i` – and squash things down

kevmoo:
@chenzhiguang – I think I fixed it.

@srawlins – PTAL

chenzhiguang:
> @chenzhiguang – I think I fixed it.
> 
> @srawlins – PTAL

Thanks @kevmoo I still trying to figure out how this `git rebase -i` works.  😞

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -71,27 +72,16 @@ class TableSyntax extends BlockSyntax {
   }
 
   List<String?> _parseAlignments(String line) {
-    final startIndex = _walkPastOpeningPipe(line);
+    final columns = line.replaceAll(RegExp(r'^\s*\||\|\s*$'), '').split('|');
```
srawlins, line 75: This feels a little fishy to me, like perhaps an unnecessary callout to the regular expression engine. I think it could be inefficient. It took me a minute to see that we're matching two patterns, `^\s*\|` and `\|\s*$`. I think we can instead use to RegExps like:

    static final p1 = RegExp(r'^\s*\|');
    static final p2 = RegExp(r'\|\s*$');
    // ...
    final columns = line.replaceFirst(p1, '').replaceFirst(p2, '').split('|');


Even beyond that, I'd want to make sure that calling out to the RegExp engine is not slower (or much slower) than the hand-written byte-walking which this replaces.

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -166,7 +156,7 @@ class TableSyntax extends BlockSyntax {
 
     for (var i = 0; i < row.length && i < alignments.length; i++) {
       if (alignments[i] == null) continue;
-      row[i].attributes['style'] = 'text-align: ${alignments[i]};';
+      row[i].attributes['align'] = '${alignments[i]}';
```
srawlins, line 159: It looks like `<td align` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#attr-align) in favor of `<td style="text-align:`.

chenzhiguang: Yes. it is. Not sure why Github spec has not updated it.

lib/src/block_syntaxes/table_syntax.dart
chenzhiguang: Yes, I doubt the performance too. Now it should be better.

lib/src/block_parser.dart
```diff
@@ -107,11 +111,25 @@ class BlockParser {
 
   List<Node> parseLines() {
     final blocks = <Node>[];
+
+    // If the `_pos` does not change before and after `parse()`, never try to
+    // parse this `_pos` with the same syntax again.
+    BlockSyntax? neverMatch;
```
srawlins, line 117: Apologies, I've read through this a few times and I can't get my head around this variable. Specifically it's name, and what the check on line 121 is for. Can you add a comment maybe on line 121 and 131. I just know that when I look at this code months from now, I'll have no idea what [neverMatch] is for.

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -71,29 +72,39 @@ class TableSyntax extends BlockSyntax {
   }
 
   List<String?> _parseAlignments(String line) {
-    final startIndex = _walkPastOpeningPipe(line);
+    final columns = <String?>[];
+    var started = false;
```
srawlins, line 76: Could you add a comment above this variable explaining when it is true and when false.

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -71,29 +72,39 @@ class TableSyntax extends BlockSyntax {
...
-        endIndex--;
-        break;
+      if (char == $colon) {
+        if (hitDash) {
+          align = align == 'left' ? 'center' : 'right';
+        } else {
+          align = 'left';
+        }
+      } else {
+        hitDash = true;
```
srawlins, line 94: Hmm, we say `hitDash = true` if `char` is not `$colon`? I don't understand, should we rename this variable?

srawlins (changes requested):

lib/src/block_parser.dart
chenzhiguang: Because of this `retreat`, which moves the pointer one step back, the result will be an infinite loop if without this `neverMatch` mechanism. 

https://github.com/dart-lang/markdown/blob/03e82eec45abd165f3959a95ef0a8722cc2bafb2/lib/src/block_syntaxes/table_syntax.dart#L39

lib/src/block_syntaxes/table_syntax.dart
chenzhiguang: Sorry, I put the hitDash at a wrong place. Fixed

lib/src/block_syntaxes/table_syntax.dart
chenzhiguang: Fixed.

lib/src/block_parser.dart
chenzhiguang: Normally in `canParse` we should make sure a structure satisfies all the requirements of the target element before ending up into `parse` method. But for some elements such as tables,  their tokens locate at the second line, they might  also relay on some other conditions or even the type of the following element to make them a valid target element. It might be low performance if we put all these validations in `canParse` method,  except we cache some calculated result from `canParse` and reuse them in `parse` method. 

With the `retreat` and `neverMatch` mechanism,  `canParse` can do a roughly match,  the `parse` has a chance to regret the `true` result from `canParse`.

I might implement the same mechanism in setext headings and link reference definitions later if it has a better performance.

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -71,29 +72,42 @@ class TableSyntax extends BlockSyntax {
   }
 
   List<String?> _parseAlignments(String line) {
-    final startIndex = _walkPastOpeningPipe(line);
+    final columns = <String?>[];
+    // Set the value to `true` when hitting a non whitespace character other
+    // than the first pipe character.
+    var started = false;
+    var hitDash = false;
+    String? align;
```
srawlins, line 80: Let's name this, `alignment`, as a noun.

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -71,29 +72,42 @@ class TableSyntax extends BlockSyntax {
...
       }
-      if (ch != $space && ch != $tab) {
-        break;
+
+      if (char == $pipe) {
+        columns.add(align);
+        hitDash = false;
+        align = null;
+      } else {
+        hitDash = true;
```
srawlins, line 102: Hmm, I think this variable maybe should be changed. So it doesn't necessarily mean, "I've hit a dash." It more means, "I've moved past the chance to hit an alignment dash." Is that right?

srawlins (changes requested):

lib/src/block_parser.dart
chenzhiguang: Thanks. I also added one more comment to this line.

lib/src/block_syntaxes/table_syntax.dart
chenzhiguang: I need something means now I am at the position of a dash ( when `final char = line.codeUnitAt(i);` is a dash)

It might not always move past a dash, it might end up on a dash when there is no ending pipe, for example
```
|--:|--
```

lib/src/block_syntaxes/table_syntax.dart
chenzhiguang: Fixed

lib/src/block_syntaxes/table_syntax.dart
chenzhiguang: The `hitDash` helps to determine if a colon is on the left `:---` or right `---:` or both side `:---:`

lib/src/block_parser.dart
```diff
@@ -107,11 +111,28 @@ class BlockParser {
 
   List<Node> parseLines() {
     final blocks = <Node>[];
+
+    // If the `_pos` does not change before and after `parse()`, never try to
+    // parse the line at `_pos` with the same syntax again.
+    // For example the `TableSyntax` might not advance the `_pos` in `parse`
+    // method, beause of the header row does not match the delimiter row in the
+    // number of cells, which makes a table like structure not be recognized.
```
srawlins, line 119: That's excellent. Super helpful comment.

lib/src/block_parser.dart
```diff
@@ -107,11 +111,28 @@ class BlockParser {
...
+    BlockSyntax? neverMatch;
+
     while (!isDone) {
       for (final syntax in blockSyntaxes) {
+        if (neverMatch != null && neverMatch == syntax) {
```
srawlins, line 124: I think we don't need to check `neverMatch != null` separate from `neverMatch == syntax` since `syntax` is never null.

lib/src/block_syntaxes/table_syntax.dart
srawlins: OK I think I understand.

srawlins (changes requested):
One little request, then we're good!

lib/src/block_parser.dart
chenzhiguang: Thanks. Fixed!

srawlins (approved):

chenzhiguang:
🎉

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -166,7 +180,7 @@ class TableSyntax extends BlockSyntax {
 
     for (var i = 0; i < row.length && i < alignments.length; i++) {
       if (alignments[i] == null) continue;
-      row[i].attributes['style'] = 'text-align: ${alignments[i]};';
+      row[i].attributes['align'] = '${alignments[i]}';
```
srawlins, line 183: Oops, I meant to press on this. Regardless of the spec, I don't think we should change this. We should stick to `<td style="text-align:" and avoid changing to a deprecated HTML attribute.

kevmoo: "deprecated" is tricky. This syntax will be supported FOR EVER. If there are zero concerns about it working, sticking to spec will save us a lot of sanity in the future IMHO

lib/src/block_syntaxes/table_syntax.dart
```diff
@@ -50,7 +51,7 @@ class TableSyntax extends BlockSyntax {
       if (children != null) {
         while (children.length < columnCount) {
           // Insert synthetic empty cells.
-          children.add(Element.empty('td'));
+          children.add(Element('td', const []));
```
chenzhiguang, line 54: Should change to:
```dart
children.add(Element('td', []));
```

kevmoo: It's a bit of a hack to support a hack, but yes.

Once the flutter markdown integration test is added, it'll be clear when this causes failures.

lib/src/block_syntaxes/table_syntax.dart
srawlins: I guess since GitHub itself uses `<td align="`, 🤷  I guess it works fine.

I was assuming this attribute perhaps had minor differences from `style="text-align:`, like different interaction with other CSS, or inheritance or something, but if that's not the case, let's go with what is specified and what GitHub actually uses.

PR #478, opened 2022-10-28
Improve FencedCodeBlockSyntax

chenzhiguang:
Strictly improves the test coverage.

### DIFF
--- a/lib/src/block_syntaxes/fenced_code_block_syntax.dart
+++ b/lib/src/block_syntaxes/fenced_code_block_syntax.dart
@@ -4,7 +4,6 @@
 
 import '../ast.dart';
 import '../block_parser.dart';
-import '../charcode.dart';
 import '../patterns.dart';
 import '../util.dart';
 import 'block_syntax.dart';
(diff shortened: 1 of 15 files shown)


kevmoo:
@chenzhiguang – please rebase on `master` – we now have `flutter_markdown` validation in place

If there are breaks there, we'll have to dig to see what's expected/intended.

chenzhiguang:
Something went wrong when rebasing, I will fix it.

chenzhiguang:
Rebased

lib/src/block_syntaxes/fenced_code_block_syntax.dart
```diff
@@ -19,77 +18,104 @@ class FencedCodeBlockSyntax extends BlockSyntax {
   const FencedCodeBlockSyntax();
 
   @override
-  bool canParse(BlockParser parser) {
```
srawlins, line 22: Can you say what effect this has, removing `canParse`? Does it mean `canParse` will always return true?

lib/src/block_syntaxes/fenced_code_block_syntax.dart
```diff
@@ -19,77 +18,104 @@ class FencedCodeBlockSyntax extends BlockSyntax {
...
-  @override
-  Node parse(BlockParser parser) {
-    // Get the syntax identifier, if there is one.
-    final match = pattern.firstMatch(parser.current)!;
-    final endBlock = match.group(1);
-    var infoString = match.group(2)!;
+class _FenceMatch {
+  _FenceMatch.fromMatch(RegExpMatch match) : indent = match[1]!.length {
+    if (match.namedGroup('backtick') != null) {
+      marker = match.namedGroup('backtick')!;
```
srawlins, line 97: Rather than initialize these `late final` fields in the constructor body, let's go ahead and make this a factory constructor, and add a `_FencedMatch._` constructor that assigns `marker` and `info` which can be `final` fields.

lib/src/block_syntaxes/fenced_code_block_syntax.dart
```diff
@@ -19,77 +18,104 @@ class FencedCodeBlockSyntax extends BlockSyntax {
...
-    final childLines = parseChildLines(parser, endBlock);
+  final int indent;
+  late final String marker;
+  late final String _info;
 
-    // The Markdown tests expect a trailing newline.
-    childLines.add('');
+  // The info-string should be trimmed,
+  // https://spec.commonmark.org/0.30/#info-string.
+  String get info => _info.trim();
```
srawlins, line 111: Rather than do this in the getter, and store an `_info` which is not trimmed, let's just do the trimming in the constructor.

srawlins (changes requested):
I think this all looks really good.

Would you mind posting benchmarks, before and after? I get ancy with RegExp changes. There are definitely strings that have really bad performance with some RegExp, and this has appeared in Flutter code before.

Just a few other requests.

chenzhiguang:
## Benchmark

### Before

Screenshot 2022-11-04 at 18 01 48

### After

Screenshot 2022-11-04 at 18 05 01

lib/src/block_syntaxes/fenced_code_block_syntax.dart
chenzhiguang: So the default `canParse` of  `BlockSyntax` applies:
```dart
bool canParse(BlockParser parser) {
  return pattern.hasMatch(parser.current);
}
```

lib/src/block_syntaxes/fenced_code_block_syntax.dart
chenzhiguang: Thanks! really cool tricks.

lib/src/block_syntaxes/fenced_code_block_syntax.dart
chenzhiguang: Fixed!

kevmoo:
@srawlins – I think benchmarks look fine, no?

chenzhiguang:
rebased

srawlins (approved):
Looks great!

chenzhiguang:
Cool, thanks!

PR #484, opened 2022-11-08
Fix HTML escape issues

chenzhiguang:
This PR introduces a new syntax `EscapeHtmlSyntax`, fixes most of the HTML escape issues.

### Before 

Screenshot 2022-11-08 at 19 17 15

### After
Screenshot 2022-11-09 at 05 00 19


chenzhiguang:
We might need to update **flutter_markdown** test cases.

kevmoo:
@chenzhiguang – let's get your other PR landed and then I can look at `flutter-markdown`

OR you could look at https://github.com/flutter/packages/pull/2777 and take inspiration.

What I did: synced down the package repo locally and just hacked with a `dependency_override` on my local pkg:markdown until tests passed with and without the override.

lib/src/inline_parser.dart
```diff
@@ -27,12 +28,6 @@ class InlineParser {
     EmailAutolinkSyntax(),
     AutolinkSyntax(),
     LineBreakSyntax(),
-    // Allow any punctuation to be escaped.
-    EscapeSyntax(),
-    // "*" surrounded by spaces is left alone.
-    TextSyntax(r' \* ', startCharacter: $space),
-    // "_" surrounded by spaces is left alone.
-    TextSyntax(' _ ', startCharacter: $space),
```
chenzhiguang, line 35: I guess these two `TextSyntax` intend to improve the performance? But the reality is opposite.

**With them:**

Screenshot 2022-11-09 at 04 37 50

**Without them**

Screenshot 2022-11-09 at 04 36 48

chenzhiguang: Maybe I should not remove them in this PR, it is not related to the topic of this PR.

kevmoo: Yeah, don't mix the "topics". A separate PR is good!

kevmoo:
@chenzhiguang – if the change IS related to fixing HTML escape, it's fine to leave, too!

chenzhiguang:
> @chenzhiguang – if the change IS related to fixing HTML escape, it's fine to leave, too!

It is not related, I will create some performance improvement PRs later.

chenzhiguang:
> What I did: synced down the package repo locally and just hacked with a `dependency_override` on my local pkg:markdown until tests passed with and without the override.

Thanks, I will do it when this PR is approved.

lib/src/util.dart
```diff
@@ -9,81 +9,24 @@ import 'assets/html_entities.dart';
...
+/// One or more whitespace, for compressing.
+final _oneOrMoreWhitespacePattern = RegExp('[ \n\r\t]+');
+
+String escapeHtml(
+  String html, {
+  bool escapeApos = true,
+  bool escapeLtGt = true,
+  bool escapeQuot = true,
+}) =>
+    HtmlEscape(HtmlEscapeMode(
```
srawlins, line 21: I think both `HtmlEscape` and `HtmlEscapeMode` can be made `const`.

srawlins (changes requested):
I've looked at about 1/3 of this 😁  sending one comment so far.

lib/src/util.dart
chenzhiguang: Thanks, fixed.

srawlins (approved):
Looks great, thanks much!

chenzhiguang:
Cool, thanks. I will update flutter_markdown later.

srawlins:
@devoncarew @kevmoo  I want to cause the least chaos this time around. Must we / can we fix flutter_markdown before landing this?

kevmoo:
We get fixes into flutter_markdown, land those, rerun the test here, then
land this PR

chenzhiguang:
The PR to fix **flutter_markdown** test cases: https://github.com/flutter/packages/pull/2797

PR #488, opened 2022-11-10
Refactor CodeBlockSyntax

chenzhiguang:
To have 100% strict.

### DIFF
--- a/lib/src/block_syntaxes/code_block_syntax.dart
+++ b/lib/src/block_syntaxes/code_block_syntax.dart
@@ -33,7 +33,7 @@ class CodeBlockSyntax extends BlockSyntax {
           pattern.hasMatch(parser.current) != true) {
         break;
       }
-      childLines.add(parser.current.indent().text);
+      childLines.add(parser.current.dedent().text);
 
       parser.advance();
     }
@@ -60,6 +60,7 @@ class CodeBlockSyntax extends BlockSyntax {
     var i = 1;
     while (true) {
       final nextLine = parser.peek(i);
+      // EOF
       if (nextLine == null) {
         return true;
       }
--- a/lib/src/block_syntaxes/code_block_syntax.dart
+++ b/lib/src/block_syntaxes/code_block_syntax.dart
@@ -23,25 +23,21 @@ class CodeBlockSyntax extends BlockSyntax {
     final childLines = <String?>[];
 
     while (!parser.isDone) {
-      final match = pattern.firstMatch(parser.current);
-      if (match != null) {
-        childLines.add(match[1]);
-        parser.advance();
-      } else {
-        // If there's a codeblock, then a newline, then a codeblock, keep the
-        // code blocks together.
-        final nextMatch =
-            parser.next != null ? pattern.firstMatch(parser.next!) : null;
-        if (parser.current.trim() == '' && nextMatch != null) {
-          childLines.add('');
-          childLines.add(nextMatch[1]);
-          parser.advance();
-          parser.advance();
-        } else {
-          break;
-        }
+      final isBlankLine = parser.current.isBlank;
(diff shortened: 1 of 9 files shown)


lib/src/util.dart
```diff
@@ -164,3 +164,55 @@ String escapePunctuation(String input) {
...
+  bool get isBlank => trim().isEmpty;
+}
+
+class IndentedText {
+  final String text;
+
+  /// How many spaces of a tab that remains after part of it has been consumed.
+  ///
+  /// `null` means it did not hit a `tab`.
+  final int? tabRemaining;
```
chenzhiguang, line 215: I will need this `tabRemaining` in another syntax.

And here is an explanation about this `tabRemaining`:

For example this an indented code block which starts with 3 spaces and a tab.

`···\t`var a = 1

A table represents 4 spaces,  one of them is consumed by the indented code block structure, so there are 3 spaces left, I call these 3 spaces `tabRemaining`

lib/src/block_syntaxes/code_block_syntax.dart
```diff
@@ -59,4 +55,23 @@ class CodeBlockSyntax extends BlockSyntax {
 
     return Element('pre', [Element.text('code', content)]);
   }
+
+  bool _shouldEnd(BlockParser parser) {
+    var i = 1;
+    while (true) {
+      final nextLine = parser.peek(i);
+      if (nextLine == null) {
```
srawlins, line 63: Can you add a comment here on what this means, (I assume it means EOF, but I'm not sure).

lib/src/util.dart
```diff
@@ -131,3 +131,55 @@ String escapePunctuation(String input) {
 
   return buffer.toString();
 }
+
+extension StringExtensions on String {
+  /// Removes leading whitespace by the length of [length].
```
srawlins, line 136: Maybe "Removes up to [length] characters of leading whitespace."

lib/src/util.dart
```diff
@@ -131,3 +131,55 @@ String escapePunctuation(String input) {
...
+  // The way of handling tabs: https://spec.commonmark.org/0.30/#tabs
+  IndentedText indent([int length = 4]) {
```
srawlins, line 138: I think this should be called _dedent_. "indent" would _add_ whitespace.

lib/src/util.dart
```diff
@@ -131,3 +131,55 @@ String escapePunctuation(String input) {
...
+      }
+    }
+    return IndentedText(substring(start), tabRemaining);
+  }
+
+  /// Whether this string contains only whitespaces.
+  bool get isBlank => trim().isEmpty;
+}
+
+class IndentedText {
```
srawlins, line 176: Please add a doc comment for this new class.

lib/src/util.dart
```diff
@@ -131,3 +131,55 @@ String escapePunctuation(String input) {
...
+  final String text;
+
+  /// How many spaces of a tab that remains after part of it has been consumed.
+  ///
+  /// `null` means it did not hit a `tab`.
```
srawlins, line 181: How about "`null` means we did not read a 'tab'`".

srawlins (changes requested):

lib/src/util.dart
chenzhiguang: Fixed. thanks!

lib/src/util.dart
chenzhiguang: Fixed

lib/src/util.dart

lib/src/util.dart
chenzhiguang: Fixed, thanks!

srawlins (approved):
Thanks

PR #489, opened 2022-11-11
Some cleanup

chenzhiguang:


lib/src/inline_parser.dart
```diff
@@ -28,10 +28,6 @@ class InlineParser {
     EmailAutolinkSyntax(),
     AutolinkSyntax(),
     LineBreakSyntax(),
-    // "*" surrounded by spaces is left alone.
-    TextSyntax(r' \* ', startCharacter: $space),
-    // "_" surrounded by spaces is left alone.
-    TextSyntax(' _ ', startCharacter: $space),
```
chenzhiguang, line 34: These two `TextSyntax` might intend to improve the performance, but the result is opposite.

srawlins (approved):

PR #491, opened 2022-11-14
benchmark: use Stopwatch over DateTime for benchmarking

kevmoo:
Use built-in String.padLeft


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

kevmoo:
Gah! I just benchmarked this – turns out the perf delta is trivial. I *think* it's better to use `Stopwatch` just because there is zero chance of timezone silly...but 🤷

srawlins (approved):

PR #494, opened 2022-11-16
introduce a Line class

chenzhiguang:
The `Line` is used to maintain the `tabRemaining` for now.

This is a use case of `Line`, it can also explain the meaning of the `tabRemaining`:
https://spec.commonmark.org/0.30/#example-6

> Normally the `>` that begins a block quote may be followed optionally by a space, which is not considered part of the content. In the following case > is followed by a tab, which is treated as if it were expanded into three spaces. Since one of these spaces is considered part of the delimiter, foo is considered to be indented six spaces inside the block quote context, so we get an indented code block starting with two spaces.

In this example the `tabRemaining` is the two spaces from **blockquote** which will be used in the nested **indented code block**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,6 @@
   line ending.
 * Add a new syntax `EscapeHtmlSyntax` to encode (`"`), (`<`), (`>`) and (`&`).
 * Add an option `caseSensitive` to `TextSyntax`.
-* **Breaking change**: Change the type of `parseLines`'s parameter from `String`
-  to `List<Line>`
 * Add a new public method `parse(String text)` for `Document`.
 
 ## 6.0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
   `indicatorForCheckedCheckBox`, and `indicatorForUncheckedCheckBox`.
 * **Breaking change**: Removed `BlockHtmlSyntax`, `BlockTagBlockHtmlSyntax`,
   `LongBlockHtmlSyntax`, and `OtherTagBlockHtmlSyntax`.
+* **Breaking change**: Change the `line` properties of type `String` to `Line`.
+* **Breaking change**: Change the `lines` properties of type `List<String>` to
+  `List<Line>`.
 * Add a new syntax `HtmlBlockSyntax` to parse HTML blocks.
 * Add a new syntax `DecodeHtmlSyntax` to decode HTML entity and numeric
   character references.
@@ -16,9 +19,6 @@
 * Add a new public method `parse(String text)` for `Document`.
 * Add a new public method `parseLineList(List<Line> text)` for `Document`.
 * Add a new type: `Line`.
-* **Breaking change**: Change the `line` properties of type `String` to `Line`.
-* **Breaking change**: Change the `lines` properties of type `List<String>` to
-  `List<Line>`.
 
 ## 6.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@
 * Add a new syntax `EscapeHtmlSyntax` to encode (`"`), (`<`), (`>`) and (`&`).
 * Add an option `caseSensitive` to `TextSyntax`.
 * Add a new public method `parse(String text)` for `Document`.
+* Add a new public method `parseLineList(List<Line> text)` for `Document`.
+* Add a new type: `Line`.
+* **Breaking change**: Change all the `String` type **_line_** to `Line` type.
+* **Breaking change**: Change all the `List<String>` type **_lines_** to
+  `List<Line>` type.
 
 ## 6.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
   line ending.
 * Add a new syntax `EscapeHtmlSyntax` to encode (`"`), (`<`), (`>`) and (`&`).
 * Add an option `caseSensitive` to `TextSyntax`.
+* **Breaking change**: Change the type of `parseLines`'s parameter from `String`
+  to `List<Line>`
+* Add a new public method `parse(String text)` for `Document`.
 
 ## 6.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,9 @@
 * Add a new public method `parse(String text)` for `Document`.
 * Add a new public method `parseLineList(List<Line> text)` for `Document`.
 * Add a new type: `Line`.
-* **Breaking change**: Change all the `String` type **_line_** to `Line` type.
-* **Breaking change**: Change all the `List<String>` type **_lines_** to
-  `List<Line>` type.
+* **Breaking change**: Change the `line` properties of type `String` to `Line`.
+* **Breaking change**: Change the `lines` properties of type `List<String>` to
+  `List<Line>`.
 
 ## 6.0.1
 
--- a/lib/markdown.dart
+++ b/lib/markdown.dart
@@ -82,5 +82,6 @@ export 'src/inline_syntaxes/link_syntax.dart';
 export 'src/inline_syntaxes/soft_line_break_syntax.dart';
 export 'src/inline_syntaxes/strikethrough_syntax.dart';
 export 'src/inline_syntaxes/text_syntax.dart';
+export 'src/line.dart';
 
 const version = packageVersion;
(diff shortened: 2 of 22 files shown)


chenzhiguang:
Can we approve this sooner?  I need it to create PRs to fix the tab related issues.

kevmoo:
We'll need to get this working w/ `flutter_markdown` – which is tricky because the API has changed!

chenzhiguang:
Yes, I think we can use the same way as we did before to fix it.

lib/src/document.dart
```diff
@@ -65,8 +67,11 @@ class Document {
     }
   }
 
+  /// Parses the given [text] to a series of AST nodes.
+  List<Node> parse(String text) => parseLines(text.toLines());
+
   /// Parses the given [lines] of Markdown to a series of AST nodes.
-  List<Node> parseLines(List<String> lines) {
+  List<Node> parseLines(List<Line> lines) {
```
kevmoo, line 74: Might want to add a NEW function here called `parseLineList` or similar. This maintains backwards compat.

Thoughts @srawlins ?

chenzhiguang: I have added  `parse(String text)`, which accepts  a string parameter.

kevmoo: We need to MAINTAIN the existing API form.

Or figure out a work-around for Flutter.

Maybe we release v7 and deprecate the current API name and add a new name. Etc etc...releasing software is complicated!

chenzhiguang: Now I got it!

chenzhiguang:
Now it looks better.

lib/src/document.dart
```diff
@@ -66,7 +68,14 @@ class Document {
...
   /// Parses the given [lines] of Markdown to a series of AST nodes.
-  List<Node> parseLines(List<String> lines) {
+  List<Node> parseLines(List<String> lines) =>
+      parseLineList(lines.map(Line.new).toList());
+
+  /// Parses the given [text] to a series of AST nodes.
+  List<Node> parse(String text) => parseLineList(text.toLines());
+
+  /// Parses the given [lines] of [Line] to a series of AST nodes.
+  List<Node> parseLineList(List<Line> lines) {
```
kevmoo, line 78: Need to document this in the changelog, too!

lib/markdown.dart
```diff
@@ -82,5 +82,6 @@ export 'src/inline_syntaxes/link_syntax.dart';
 export 'src/inline_syntaxes/soft_line_break_syntax.dart';
 export 'src/inline_syntaxes/strikethrough_syntax.dart';
 export 'src/inline_syntaxes/text_syntax.dart';
+export 'src/line.dart';
```
kevmoo, line 85: changelog: document new type, too!

Also: we changed the public API of public classes here, right? `lines` field now different. should be in changelog, too!

lib/src/document.dart
chenzhiguang: Fixed!

lib/markdown.dart
chenzhiguang: Fixed!

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@
 * Add a new syntax `EscapeHtmlSyntax` to encode (`"`), (`<`), (`>`) and (`&`).
 * Add an option `caseSensitive` to `TextSyntax`.
 * Add a new public method `parse(String text)` for `Document`.
+* Add a new public method `parseLineList(List<Line> text)` for `Document`.
+* Add a new type: `Line`.
+* **Breaking change**: Change all the `String` type **_line_** to `Line` type.
```
kevmoo, line 19:
```suggestion
* **Breaking change**: Change the **_line_** properties of type `String` to `Line`.
```

and similar below!

chenzhiguang: Done

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@
   line ending.
 * Add a new syntax `EscapeHtmlSyntax` to encode (`"`), (`<`), (`>`) and (`&`).
 * Add an option `caseSensitive` to `TextSyntax`.
+* Add a new public method `parse(String text)` for `Document`.
+* Add a new public method `parseLineList(List<Line> text)` for `Document`.
+* Add a new type: `Line`.
+* **Breaking change**: Change the `line` properties of type `String` to `Line`.
```
kevmoo, line 19: one small nit: put all breaks at the top together!

chenzhiguang: Done

kevmoo:
Help me understand the *point* of `Line`? The API is now a bit more complex. What do we get by having a `Line` construct instead of just using `String`? (Might be good to have in the change log, too)

chenzhiguang:
I found It seems the tab issues could be fixed without using this `Line` type. I will do more research tomorrow. 

The idea came from https://github.com/tagnote-app/dart_markdown/blob/master/lib/src/line.dart, but the Line there maintains much more information.

kevmoo:
Also: mind running benchmarks?

chenzhiguang:
No difference, I have tried. Let me do it again

kevmoo:
just because we're allocating a bit more...

chenzhiguang:
### Benchmark before

Screenshot 2022-11-16 at 21 59 12

### Benchmark after

Screenshot 2022-11-16 at 21 58 29

chenzhiguang:
We might need this `Line` in the future.

chenzhiguang:
@kevmoo 
We might still need `Line` to maintain the tab `tabRemaining` we can not just simply convert the `tabRemaining` to spaces and prepend them to a string, then passes it to the nested structure.

For example: (See: https://spec.commonmark.org/0.30/#example-5)

```markdown
- foo

→→bar
```

There are 2 space sized `tabRemaining` from the list block, if we just pretend two spaces to `→bar` like this `..→bar`, 
the output will be: (See: https://spec.commonmark.org/0.30/#example-2)
```html
<pre><code>foo
</code></pre>
```

chenzhiguang:
I have added some more comment for the `tabRemaining` property of `Line`.

Sorry for this back and forth, it has been a few months since I created this `Line` type, I should have written a detailed comment.

chenzhiguang:
Should we continue? I am blocked by this PR.

kevmoo:
@srawlins should really chime in here

chenzhiguang:
What do we say?

chenzhiguang:
@srawlins Hi Sam,  I need your decision in order to plan the next steps.
cc @kevmoo

srawlins:
I will try to get to this today.

chenzhiguang:
@srawlins Hi Sam, please don’t forget this one.

lib/src/line.dart
```diff
@@ -0,0 +1,53 @@
...
+  // ```html
+  // <pre><code>  foo
+  // </code></pre>
+  // ```
+  final int? tabRemaining;
+
+  // A line containing no characters, or a line containing only spaces
+  // (`U+0020`) or tabs (`U+0009`), is called a blank line.
+  // https://spec.commonmark.org/0.30/#blank-line
+  bool get isBlankLine => emptyPattern.hasMatch(content);
```
srawlins, line 41: It'd be interesting to consider how many `Line`s are queried for whether they are blank. I suspect it is a high number, but it'd be good to really look. And how many `Line`s are queried multiple times whether they are blank. If most `isBlankLine` is called on most `Line`s, and called multiple times on some `Line`s, it might be a win to store this `bool` as a final field.

lib/src/line.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  const Line(
+    this.content, {
+    this.tabRemaining,
+  });
+
+  Map<String, dynamic> toMap() => {
```
srawlins, line 48: If this is only called in tests, I don't think it is worth it to implement here.

test/util_test.dart
```diff
@@ -52,3 +53,11 @@ void main() {
     });
   });
 }
+
+extension LineX on Line {
```
kevmoo, line 57: If you only use it in this file, you can drop the `LineX` name.
```suggestion
extension on Line {
```

chenzhiguang: Thanks, I did not even know that 😄

lib/src/line.dart
chenzhiguang: Fixed. The ListSyntax does call `isBlankLine` on most lines and also calls multiple times on some lines

test/util_test.dart
srawlins: Great solution to make this a testing function

srawlins (approved):
Super cool, great feature, sorry for the delay.

chenzhiguang:
No worries! Thanks!

Issue #497, opened 2022-11-29
Single quote turning into &#39; in fenced code block


derekxu16:
I believe that only `&`, `<`, and `>` should turn into HTML entities when they're within fenced code blocks; `'` should not.

The following is a way to reproduce this problem:

```
import 'dart:developer';
import 'package:markdown/markdown.dart' show Document;

void main(List<String> args) async {
  final document = Document();
  final nodes = document.parseLines(['```', "'abc'", '```']);
  debugger();
  print(nodes);
}
```

Inspecting the `nodes` array in this example reveals the problem.
<img width="348" alt="Screen Shot 2022-11-29 at 11 16 57 AM" src="https://user-images.githubusercontent.com/1782969/204584284-df24cb74-2156-4cf2-ba40-55798225a07f.png">

srawlins:
CC @chenzhiguang I believe this is intentional, according to CommonMark spec. I'm not certain though.

chenzhiguang:
I can't find any information on this in the spec, but I could see the CommonMark live demo does not encode `'`:  https://spec.commonmark.org/dingus/?text=%60%60%60%0A%27abc%27%0A%60%60%60%, we should not encode `'`. 

I will fix it later soon.

PR #498, opened 2022-11-29
Do not escape single quote(apostrophe) inside code

chenzhiguang:
Fix: https://github.com/dart-lang/markdown/issues/497


srawlins (approved):
Thanks!

PR #506, opened 2023-01-14
Rewrite link reference definitions

chenzhiguang:
Now it has a clean `ParagraphSyntax`.

### Before
Screenshot 2023-01-14 at 16 35 52

### After
Screenshot 2023-01-14 at 16 34 53


lib/src/link_parser.dart
```diff
@@ -0,0 +1,265 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'charcode.dart';
+import 'text_parser.dart';
+import 'util.dart';
+
+class LinkParser extends TextParser {
```
chenzhiguang, line 9: I will implement this `LinkParser` to `LinkSyntax`, this is the reason why I create it separately.

lib/src/link_parser.dart
```diff
@@ -0,0 +1,265 @@
...
+  /// If there is a valid link formed.
+  bool valid = false;
```
kevmoo, line 11: If these fields are not used by other types, please make them private!

kevmoo: Seeing that they are used, then they should be getters. I know it's extra work, but we want to avoid exposing unwanted mutable state

chenzhiguang:
@srawlins Hi Sam, please do not forget this one when you have time. And there are some more PRs after it. 🙂

kevmoo:
@chenzhiguang – just updated your branch w/ my latest change

chenzhiguang:
> @chenzhiguang – just updated your branch w/ my latest change

Thanks.  I am just about to do it.

srawlins (approved):
100% commonmark 🎉 . I love it. No comments.

chenzhiguang:
Awesome! 🥳

PR #513, opened 2023-02-01
Prepend a line ending if the p tag is removed and there is an element before it in a *tight* list

chenzhiguang:


srawlins (approved):
LGTM!

PR #517, opened 2023-02-14
blast_repo fixes

kevmoo:
auto-publish
github-actions
no-response

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,14 @@
+# A CI configuration to auto-publish pub packages.
+
+name: Publish
+
+on:
+  pull_request:
+    branches: [ master ]
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+on:
+  pull_request:
+    branches: [ master ]
+  push:
+    tags: [ 'v[0-9]+.[0-9]+.[0-9]+*' ]
+
+jobs:
+  publish:
+    if: ${{ github.repository_owner == 'dart-lang' }}
```
devoncarew, line 13: 👍 

we'll also need to update the admin page at pub.dev

devoncarew (approved):

.github/workflows/publish.yaml
kevmoo: Done!

Issue #521, opened 2023-02-16
Exception on parsing # characters with GitHubWeb extension (markdown 7.0.1)


isoos:
Original markdown content from https://pub.dev/packages/hycop/versions/0.1.0

The exception is thrown on the first line of this input:
```
# ############################
# Run
# ############################
```

```
dart run markdown --extension-set GitHubWeb <x.md

Unhandled exception:
Bad state: No element
#0      List.first (dart:core-patch/growable_array.dart:343:5)
#1      BlockSyntax.generateAnchorHash (package:markdown/src/block_syntaxes/block_syntax.dart:59:25)
#2      HeaderWithIdSyntax.parse (package:markdown/src/block_syntaxes/header_with_id_syntax.dart:17:39)
#3      BlockParser.parseLines (package:markdown/src/block_parser.dart:177:32)
#4      Document.parseLineList (package:markdown/src/document.dart:79:44)
#5      Document.parse (package:markdown/src/document.dart:75:36)
#6      markdownToHtml (package:markdown/src/html_renderer.dart:40:26)
#7      main (file:///home/isoos/.pub-cache/hosted/pub.dev/markdown-7.0.0/bin/markdown.dart:67:9)
#8      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:295:33)
#9      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192:26)
```

chenzhiguang:
Here is the fix: https://github.com/dart-lang/markdown/pull/522

jonasfj:
@chenzhiguang, @srawlins can we do a patch release with this?

PR #526, opened 2023-03-29
Prepare a release of 7.0.2

jonasfj:
We'd love the fix from https://github.com/dart-lang/markdown/issues/521 for pub.dev


isoos:
Note: the bug is blocking the upgrade of dartdoc, which is also needed for new Dart language features.

srawlins:
I believe you need to bump lib/src/version.dart as well.

Issue #539, opened 2023-05-03
Table display fails if there are trailing spaces


jgfoster:
For example, see https://github.com/ponnamkarthik/FlutterToast/issues/435. Or try the following in https://dart-lang.github.io/markdown/. 

```
### good

| property   | description                               | default    |  
| ---------- | ----------------------------------------- |------------|
| child      | Widget (Not Null)(required)               | required   |  

### bad
  
| property   | description                               | default    |  
| ---------- | ----------------------------------------- |------------|  
| child      | Widget (Not Null)(required)               | required   |  
```

chenzhiguang:
Thanks! I can reproduce it.  I will fix it asap.

chenzhiguang:
@jgfoster Thank you for your interest in contributing to our project. While the issue you mentioned may be relatively simple to fix, we do think that it requires a better understanding of our project as a whole. It may be more ideal to create a dedicated project for your students to practice the Git cycle. We wish you the best of luck with your class!

PR #540, opened 2023-05-04
Fix delimiter row matching pattern for tables

chenzhiguang:
Fix: https://github.com/dart-lang/markdown/issues/539


srawlins (approved):
Thanks!

PR #541, opened 2023-05-04
wip

kevmoo:


kevmoo:
Looks like GFM got rid of their nested `<strong>` tags, which is nice.

Care to ponder a fix @chenzhiguang?

kevmoo:
Or not. Seems like this spec is diverging from the common-mark spec 🤷

Issue #543, opened 2023-05-19
paragraph + table is not rendered separately


isoos:
- `version: 7.1.0`
- original reported: https://github.com/dart-lang/pub-dev/issues/6674

The base table rendering works:
```
 dart run markdown --extension-set GitHubWeb

| `dfn <command>`                   | description                                         |
| --------------------------------- | --------------------------------------------------- |
| `dfn`                             | core tool                                           |
| `dfn <script>`                    | run a registered script                             |

<table>
<thead>
<tr>
<th><code>dfn &lt;command&gt;</code></th>
<th>description</th>
</tr>
</thead>
<tbody>
<tr>
<td><code>dfn</code></td>
<td>core tool</td>
</tr>
<tr>
<td><code>dfn &lt;script&gt;</code></td>
<td>run a registered script</td>
</tr>
</tbody>
</table>
```

However, if we put a paragraph right before the table, the table is treated as part of the paragraph:

```
x
| `dfn <command>`                   | description                                         |
| --------------------------------- | --------------------------------------------------- |
| `dfn`                             | core tool                                           |
| `dfn <script>`                    | run a registered script                             |

<p>x
| <code>dfn &lt;command&gt;</code>                   | description                                         |
| --------------------------------- | --------------------------------------------------- |
| <code>dfn</code>                             | core tool                                           |
| <code>dfn &lt;script&gt;</code>                    | run a registered script                             |</p>
```

isoos:
@srawlins @chenzhiguang any estimate when the next version will get published with this fix?

PR #545, opened 2023-05-20
Table should be able to interrupt other blocks

chenzhiguang:
- Make table be able to interrupt other blocks, such as paragraph, fix: https://github.com/dart-lang/markdown/issues/543

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


srawlins (approved):
Thanks for the fix!

PR #576, opened 2024-01-19
address a termination issue with GitHub alert syntax parsing

devoncarew:
- address a termination issue with GitHub alert syntax parsing

When trying out the new alert syntax, it looks like the parsing never terminates - all the content after an alert is included in the alert.

cc @AlexV525

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

### DIFF
--- a/lib/src/block_syntaxes/alert_block_syntax.dart
+++ b/lib/src/block_syntaxes/alert_block_syntax.dart
@@ -25,9 +25,10 @@ class AlertBlockSyntax extends BlockSyntax {
         parser.lines.any((line) => _contentLineRegExp.hasMatch(line.content));
   }
 
-  // Whether this alert ends with a lazy continuation line.
-  // The definition of lazy continuation lines:
-  // https://spec.commonmark.org/0.30/#lazy-continuation-line
+  /// Whether this alert ends with a lazy continuation line.
+  ///
+  /// The definition of lazy continuation lines:
+  /// https://spec.commonmark.org/0.30/#lazy-continuation-line
   static bool _lazyContinuation = false;
   static final _contentLineRegExp = RegExp(r'>?\s?(.*)*');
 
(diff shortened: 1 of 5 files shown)


lib/src/block_syntaxes/alert_block_syntax.dart
```diff
@@ -25,7 +25,7 @@ class AlertBlockSyntax extends BlockSyntax {
         parser.lines.any((line) => _contentLineRegExp.hasMatch(line.content));
   }
 
-  /// Whether this alert ends with a lazy continuation line.
+  // Whether this alert ends with a lazy continuation line.
```
srawlins, line 28: nit: Actually I think all three lines here should use `///`, documenting the member.

srawlins (approved):
These test cases look right to me; thanks!

lib/src/block_syntaxes/alert_block_syntax.dart
parlough: Personally, if it explains how a method is used rather than implementation details, I still use doc comments for private members as then the analysis server can surface them on hover :)

AlexV525:
@devoncarew Thanks for the fix!

devoncarew:
For sure! I think there's one remaining - though much more minor - issue. You can't currently have alerts w/ multiple paragraphs (filed as #577).

PR #582, opened 2024-02-14
Migrate example to pkg:web, update minimum required Dart version

kevmoo:


srawlins (approved):
Looks pretty mechanical, thanks!

Issue #584, opened 2024-02-22
Crash "Bad state: No element" in AlertBlockSyntax.parseChildLines


jonasfj:
I ran `test/crash_test.dart` and found 9 markdown files where `package:markdown` crashes:

```
Bad state: No element
#0      List.last (dart:core-patch/growable_array.dart:348:5)
#1      AlertBlockSyntax.parseChildLines (package:markdown/src/block_syntaxes/alert_block_syntax.dart:54:35)
#2      AlertBlockSyntax.parse (package:markdown/src/block_syntaxes/alert_block_syntax.dart:85:24)
#3      BlockParser.parseLines (package:markdown/src/block_parser.dart:178:32)
#4      Document.parseLineList (package:markdown/src/document.dart:87:44)
#5      Document.parse (package:markdown/src/document.dart:83:36)
#6      markdownToHtml (package:markdown/src/html_renderer.dart:40:26)
```

Full log in: https://gist.github.com/jonasfj/9d8dfb702f46e64b78a22f5762ac2f95

This ran a scan of all `.md` files in the latest version of all packages on pub.dev.

--------------

@AlexV525, @devoncarew, could this be related to:
 * https://github.com/dart-lang/markdown/pull/570, or
 * https://github.com/dart-lang/markdown/pull/576

AlexV525:
@jonasfj Thanks for the report. Unfortunately, I'm doing other work which occupied most of my time so I cannot respond to this in recent days. I can look into this when I'm available.

PR #585, opened 2024-02-22
Implement `AlertBlockSyntax` using `BlockquoteSyntax`

jonasfj:
A proposed fix for https://github.com/dart-lang/markdown/issues/584

As I understand the alert-block syntax, it is an extension on the original blockquote support in markdown. Basically, it's a blockquote that is rendered differently.

That means that if alert-block syntax isn't supported a markdown parser will render it as a blockquote. This provides a nice graceful fallback. It also means that if you're implementing it, you really should implement it the same way you implement blockquote parsing.

This is an attempt at implementing `AlertBlockSyntax` using the same parsing logic as `BlockquoteSyntax`.

**This is not pretty.** Ideally, whether to support alert-syntax should be a boolean option passed to `BlockquoteSyntax`. Because an alert-syntax is a blockquote with extra bling. But that's not how this package is structured :rofl: 

In fact, it can probably be argued that trying to make a markdown parser that consists of composeable objects is extremely sketchy. Especially, when we let consumers of the package implement their own syntaxes and inject them into the whole mix. It's very hard to determine how this whole thing works (and what combinations of syntaxes will work). It would probably have been safer to pass in all the option as boolean configuration options that enable/disable a particular syntaxes in a single large parser.


jonasfj:
Currently, running crash-test on this, though that won't reveal if we have bugs. Just that it doesn't crash :rofl:

jonasfj:
No crashes scanning all markdown files in the latest version for all packages on pub.dev
```
Scanned 49735 / 49913 (skipped 0), found 0 issues
## Finished scanning
Scanned 49913 package versions in 0:23:20.241887
23:21 +1: All tests passed!
```

lib/src/block_syntaxes/alert_block_syntax.dart
```diff
@@ -1,110 +0,0 @@
...
-          ? null
-          : _contentLineRegExp.firstMatch(strippedContent);
-      if (match != null) {
-        childLines.add(Line(strippedContent));
-        parser.advance();
-        _lazyContinuation = false;
-        continue;
-      }
-
-      final lastLine = childLines.last;
```
jonasfj, line 54: `childLines` can be empty at this point.

lib/src/block_syntaxes/alert_block_syntax.dart
```diff
@@ -1,110 +0,0 @@
...
-    // Grab all of the lines that form the alert, stripping off the ">".
-    final childLines = <Line>[];
-    _lazyContinuation = false;
-
-    while (!parser.isDone) {
-      final strippedContent =
-          parser.current.content.replaceFirst(RegExp(r'^\s*>?\s*'), '');
-      final match = strippedContent.isEmpty
-          ? null
-          : _contentLineRegExp.firstMatch(strippedContent);
```
jonasfj, line 46: This makes it unable to distinguish between an empty line and a line containing only `>`, but these are very different at result in different behavior.

lib/src/block_syntaxes/blockquote_syntax.dart
```diff
@@ -97,3 +97,65 @@ class BlockquoteSyntax extends BlockSyntax {
...
+    final children = BlockParser(childLines, parser.document).parseLines(
+      // The setext heading underline cannot be a lazy continuation line in a
+      // block quote.
+      // https://spec.commonmark.org/0.30/#example-93
+      disabledSetextHeading: BlockquoteSyntax._lazyContinuation,
+      parentSyntax: this,
+    );
+
+    if (isBlockquote) {
+      return Element('blockquote', children);
```
jonasfj, line 143: This is ugly, but I don't see a better way, unless we make `BlockquoteSyntax` do everything.

I suppose we could in theory make `BlockquoteSyntax` check if `AlertBlockSyntax` is enabled by doing:`

```dart
final isAlertSyntaxEnabled = parser.document.blockSyntaxes.any(
  (s) => s is AlertBlockSyntax,
);
```

If we could give `AlertBlockSyntax` a factory singleton constructor we could do:

```dart
final isAlertSyntaxEnabled = parser.document.blockSyntaxes.contains(
  const AlertBlockSyntax(),
);
```

If inside `BlockquoteSyntax` we had `isAlertSyntaxEnabled` then it could implement both. Ofcourse this would mean that enabling `AlertBlockSyntax` wouldn't work without enabling `BlockquoteSyntax`.
I'm not sure why we should ever allow such configuration, we surely don't intent to test it, but that's how the package is structured.


---------------------

The point with implementing alert-syntax inside `BlockquoteSyntax` is that alert-syntax is really a post-processing step for blockquote. But I don't really see how `package:markdown` has the facilities to post-processing.

lib/src/block_syntaxes/blockquote_syntax.dart
```diff
@@ -97,3 +97,65 @@ class BlockquoteSyntax extends BlockSyntax {
...
+  Node parse(BlockParser parser) {
+    // Parse the alert type from the first line.
+    final type =
+        pattern.firstMatch(parser.current.content)!.group(1)!.toLowerCase();
+
+    final childLines = parseChildLines(parser);
+    // Until we've parse all the child lines, we can't actually know if this is
+    // a blockquote containing `[!note]` or if this is an alert-block.
+    //
+    // This is because `> [!note]` is not a valid alert-block!
```
sigurdm, line 125: Can you explain **why** this is not a valid alert block? because it is empty?

jonasfj: I don't know, that's how it behaves on github :D

sigurdm: But is it the emptiness that makes it invalid, or some other property?

jonasfj: # Example of the empty case

```md
-----------
> [!note]
-----------
```

-----------
> [!note]
-----------


# Example of the non-empty case

```md
-----------
> [!note]
> Hello world
-----------
```

-----------
> [!note]
> Hello world
-----------

# Conclusion

If we have `> [!note]` it's not a special block. But if it's:
 * Followed by text that wants to belong to the previous paragraph, then it is. OR
 * Followed by lines like `> ...` indicating that whatever is there belongs inside the blockquote, then it also is special.

Otherwise, it's not special, just a normal blockquote.

Notice that `> foo\nbar` is the same as `> foo\n> bar`, the line that follows is considered to be a paragraph continuation of the previous line, so the blockquote continues.

So yes, as far as I can see, it's emptiness that decides it. Even if you have `> [!note]\n> ` it'll still not be rendered as a special alert-block.

lib/src/block_syntaxes/alert_block_syntax.dart
```diff
@@ -1,110 +0,0 @@
...
-class AlertBlockSyntax extends BlockSyntax {
-  const AlertBlockSyntax();
-
-  @override
-  RegExp get pattern => alertPattern;
-
-  @override
-  bool canParse(BlockParser parser) {
-    return pattern.hasMatch(parser.current.content) &&
-        parser.lines.any((line) => _contentLineRegExp.hasMatch(line.content));
```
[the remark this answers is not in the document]
devoncarew: > I suspect that parser.lines is the full text document.

Hmm, that might explain https://github.com/dart-lang/markdown/issues/579 - parsing times being quadratic.

srawlins:
I've just landed https://github.com/dart-lang/markdown/pull/593 as a fix for the crash, but I think this PR is probably going in a direction that would be good in the long term, so I'd like to keep it open. I'll try to review more tomorrow.

I looked at it today, and the only think I don't like is the same thing you don't like @jonasfj haha, which you elaborate on above.

I think there should be a way to "nicely" sit this syntax into the GFM extension, "on top" of the blockquote syntax.

> In fact, it can probably be argued that trying to make a markdown parser that consists of composeable objects is extremely sketchy.

Haha this is really true. It's a weird concept, to have a plugable parser like this, based on an "ordering" of syntaxes, where we ask each in an order, "Can you parse this line? No? And you?" It's inherently flawed.

jonasfj:
> Haha this is really true. It's a weird concept, to have a plugable parser like this, based on an "ordering" of syntaxes, where we ask each in an order, "Can you parse this line? No? And you?" It's inherently flawed.

It's completely orthogonal, maybe it deserves a controversial proposal, where we can shoot down the idea :D

Filed https://github.com/dart-lang/markdown/issues/594

mosuem:
Closing as the [dart-lang/markdown](https://github.com/dart-lang/markdown) repository is merged into the [dart-lang/tools](https://github.com/dart-lang/tools) monorepo. Please re-open this PR there!

PR #590, opened 2024-02-23
Run a periodic crash-test

jonasfj:
* Rewrote `crash_test.dart` so that:
    * It runs markdown in an isolate, insuring that infinite-loops are handled.
    * Find all markdown files using a list of file extensions.
 * Setup a periodic Github Action to run `crash_test.dart` weekly.

The `crash_test.dart` concept is a bit dumb, but it has found bugs a few times. This will download latest version of packages on pub.dev, on my desktop that takes about 25 minutes. I'm guessing that on Github Actions it'll run a bit slower :see_no_evil: 

We could run it for all package versions, but there is like 10x more package versions than packages. It's more likely that we should run it for most packages or something else.

Or perhaps we can find a nice large corpus of markdown files somewhere else to use for crash-testing.

### DIFF
--- a/test/crash_test.dart
+++ b/test/crash_test.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 import 'dart:io';
+import 'dart:isolate';
 
 import 'package:http/http.dart' as http;
 import 'package:http/retry.dart' as http;
@@ -13,6 +14,18 @@ import 'package:tar/tar.dart';
 import 'package:test/test.dart';
 
 // ignore_for_file: avoid_dynamic_calls
+
+const extensions = [
+  '.md',
+  '.mkd',
+  '.mdwn',
+  '.mdown',
+  '.mdtxt',
+  '.mdtext',
+  '.markdown',
+  'README',
+  'CHANGELOG',
+];
 
 void main() async {
   // This test is a really dumb and very slow crash-test.
@@ -26,6 +39,16 @@ void main() async {
   test(
     'crash test',
     () async {
+      final started = DateTime.now();
+      var lastStatus = DateTime(0);
+      void status(String Function() message) {
+        if (DateTime.now().difference(lastStatus) >
+            const Duration(seconds: 30)) {
+          lastStatus = DateTime.now();
+          print(message());
+        }
+      }
+
       final c = http.RetryClient(http.Client());
       Future<dynamic> getJson(String url) async {
         final u = Uri.tryParse(url);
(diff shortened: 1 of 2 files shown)


test/crash_test.dart
```diff
@@ -50,79 +73,158 @@ void main() async {
...
+          if (entry != null) {
+            packageVersions.add(PackageVersion(
+              package: package,
+              version: entry['version'] as String,
+              archiveUrl: entry['archive_url'] as String,
+            ));
+          }
+
+          // To analyze all package versions use the following lines.
+          // beware that this extremely, and probably scans that same files
```
srawlins, line 97: You've got me waiting on pins and needles, "beware that this is extremely"... what??

Also, nit: start with a capital.

test/crash_test.dart
```diff
@@ -50,79 +73,158 @@ void main() async {
...
+          // many times for no good reason.
+          /*packageVersions.addAll(
```
srawlins, line 99: Commented out lines should not be checked in; they will bit rot, and no one will how to recreate the intenion months after they become dead.

Instead, they can be wrapped with something like `if (shouldAddAllVersions)` and a flag above, `const shouldAddAllVersions = false;`

test/crash_test.dart
```diff
@@ -50,79 +73,158 @@ void main() async {
...
-            });
-          } on FormatException {
+
+          final result = await _findMarkdownIssues(
+            pv.package,
+            pv.version,
+            archive,
+          );
+
+          // If tar decoding fails
```
srawlins, line 143: Nit: end with a period, or colon. https://dart.dev/effective-dart/documentation#do-format-comments-like-sentences

test/crash_test.dart
```diff
@@ -50,79 +73,158 @@ void main() async {
...
+          '${DateTime.now().difference(started)}');
+
       if (errors.isNotEmpty) {
         print('Found issues:');
         errors.forEach(print);
         fail('Found ${errors.length} cases where markdownToHtml threw!');
       }
     },
-    timeout: const Timeout(Duration(hours: 1)),
+    timeout: const Timeout(Duration(hours: 5)),
```
srawlins, line 171: Woof, lol. Quite a task :D

test/crash_test.dart
```diff
@@ -50,79 +73,158 @@ void main() async {
...
+  final String archiveUrl;
+
+  PackageVersion({
+    required this.package,
+    required this.version,
+    required this.archiveUrl,
+  });
+}
+
+Future<List<String>?> _findMarkdownIssues(
```
srawlins, line 188: Could you add a comment here noting what it means when this returns a Future of `null`.

test/crash_test.dart
```diff
@@ -50,79 +73,158 @@ void main() async {
...
+                'package:$package-$version/${entry.name} took $time to process');
+          }
+        }
+      });
+      return issues;
+    } on FormatException {
+      return null;
+    }
+  }).timeout(const Duration(minutes: 2), onTimeout: () {
+    return ['package:$package-$version failed to be processed in 30s'];
```
srawlins, line 228: Hmm, should this say "2 minutes"?

srawlins (changes requested):
Very cool, thanks much!

test/crash_test.dart
jonasfj: Yeah, Github Actions might be slow, and we're just parsing a LOT of markdown files :rofl:

test/crash_test.dart
jonasfj: Let's just drop it, it'll take 10x longer, and if we really wanted a larger corpus of markdown files we are better of finding different ones.

srawlins (approved):
Love this so much. Thanks @jonasfj !! Everything looks great.

